PR #3, opened 2018-10-24
Kick off idle connections

an-tao:


an-tao:
Kick off idle connections

PR #30, opened 2019-05-25
Fixed std::future_error on startup

VayuDev:
Fixed std::future_error with the message "no valid state" (or similar) being randomly thrown on startup of drogon in EventLoopThread.cc in line 62. This was caused by the separate thread being started before the whole EventLoopThread object was fully constructed. As I seem to be one of the few affected ones, here is my system:
Fedora 29, GCC 8.3.1, Intel Celeron N3450 (quite slow, maybe that's related to the object being constructed too slow).
I started my drogon webserver since this change about 50 times and haven't had a single crash since then, so it's safe to assume that I fixed the problem after many, many hours. If you need, I can post my drogon-code here as well.

### DIFF
--- a/trantor/net/EventLoopThread.h
+++ b/trantor/net/EventLoopThread.h
@@ -40,11 +40,11 @@ class EventLoopThread : NonCopyable
     EventLoop *_loop;
     std::string _loopThreadName;
     void loopFuncs();
-    std::thread _thread;
     std::promise<EventLoop *> _promiseForLoopPointer;
     std::promise<int> _promiseForRun;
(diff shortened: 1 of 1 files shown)


an-tao:
After some investigation, I think you are right. There is a race condition. If the std::thread object runs to line 62 of EventLoopThread.cc before the `_promiseForRun` construct is completed, the error you find will appear. Moving it to the end of the member list ensures that the promise object is constructed before the thread runs.
I have not found this error before, maybe this error is not easy to happen in my environment, but it is indeed a problem.
Thank you for your repair!

PR #40, opened 2019-07-20
Remove the std::any

an-tao:


CMakeLists.txt
```diff
@@ -79,27 +81,15 @@ ENDIF (CMAKE_SYSTEM_NAME MATCHES "Linux")
 
 SET(CONFIG_HEADER "${PROJECT_SOURCE_DIR}/config.h")
```
OvermindDL1: All of this CONFIG_HEADER stuff really should not exist.  CMake templates should be used if you need something like this, but the define's I'm not sure why they exist, why do you need to do things like `#define CSS_STD 14` in the source, source code should not depend on C++ versions like this, rather feature tests should be done and explicit features should be enabled or disabled, those saved for install via a cmake template (if necessary at all, if not header-only that is).

trantor/net/TcpConnection.h
```diff
@@ -58,20 +58,49 @@ class TcpConnection
     virtual void shutdown() = 0;
     virtual void forceClose() = 0;
     virtual EventLoop *getLoop() = 0;
+    template <typename T>
```
OvermindDL1: :+1:

trantor/utils/LogStream.h
```diff
@@ -21,7 +21,9 @@
 #include <assert.h>
 #include <string.h>  // memcpy
 #include <string>
-
+#if CXX_STD >= 17
```
OvermindDL1: Only adding things to the API depending on the something is fine, but it should not be something like a C++ version.  Rather it should just be a normal define, not in a config file or anything, especially as these are header-inlined, that way the user can opt to compile it in or out on their choice.  What if they are using a third-party string_view, such as boost's or one of the others?  Would be good to have the define's based on that where the user can state it as they need, no config file definitions needed.  In addition this will fail to compile if the library is compiled with C++17 but the user library is compiled with C++14 (very real possibility), which is why such define's should not be baked into a config header.

CMakeLists.txt
```diff
@@ -1,28 +1,35 @@
...
+        set(CMAKE_CXX_STANDARD 14)
     endif()
 endif()
 
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+if(CMAKE_CXX_STANDARD LESS 14)
+    message(FATAL_ERROR "Error: c++14 or c++17 is required!")
+endif()
```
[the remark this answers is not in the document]
an-tao: I am afraid not, because I still need std::string_view, as belows:
https://github.com/an-tao/trantor/blob/931ada067572ed9e2ed5a2197267eec3c8c25d1a/trantor/utils/LogStream.h#L195-L204

CMakeLists.txt
```diff
@@ -64,12 +67,11 @@ IF (CMAKE_SYSTEM_NAME MATCHES "Linux")
     ADD_LIBRARY(trantor ${DIR_SRC} ${PROJECT_SOURCE_DIR}/trantor/net/inner/poller/EpollPoller.cc)
     MESSAGE(STATUS "current platform: Linux ")
 ELSEIF (CMAKE_SYSTEM_NAME MATCHES "Windows")
-    MESSAGE(STATUS "current platform: Windows")
+    MESSAGE(FATAL_ERROR "Error: Currently does not support Windows")
```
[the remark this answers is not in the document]
an-tao: In order to support windows, there is still a lot of work to be done, I plan to use IOCP to support windows in the future.

trantor/utils/LogStream.h
an-tao: How about changing the line to `#ifdef USE_STD_STRING_VIEW`(If users want to enable the API, they should use cmake like `cmake .. -DUSE_STD_STRING_VIEW`)?

If the user library is compiled with C++14, they must use `cmake .. -DCMAKE_CXX_STANDARD=14` to compile the trantor.

I seem to understand some of your meaning, the trantor must now be distributed in the source code (compiled and installed by the user), if I modify it according to your principles, trantor seems to be available in the form of a library in addition to the source distribution (Header and library files), is my understanding correct?

an-tao:
@OvermindDL1 Thank you very much for your suggestions, I will take some time to modify the PR.

CMakeLists.txt
```diff
@@ -54,7 +57,7 @@ AUX_SOURCE_DIRECTORY(trantor/net/inner DIR_SRC)
 find_package (OpenSSL)
 if(OpenSSL_FOUND)
     #add_definitions(-DUSE_OPENSSL)
-    set(DEFS "USE_OPENSSL;${DEFS}")
+    set(definitions "USE_OPENSSL;${definitions}")
     include_directories(${OPENSSL_INCLUDE_DIR})
     link_libraries(${OPENSSL_LIBRARIES})
     AUX_SOURCE_DIRECTORY(trantor/net/ssl DIR_SRC)
```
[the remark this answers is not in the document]
an-tao: When I used `find_package(OpenSSL CONFIG)`, I got an error on both MacOS and Linux:
```
Could NOT find OpenSSL (missing: OpenSSL_DIR)
``` 
And if I remove the `USE_OPENSSL` macro, How do I handle conditional compilation like this？
https://github.com/an-tao/trantor/blob/ad24f2c6733681f69c0c5bd9f218ef35c4f346d2/trantor/net/TcpServer.cc#L172-L193

trantor/utils/LogStream.h
an-tao: Can I use the __cplusplus macro?

CMakeLists.txt
an-tao: I've removed the CONFIG_HEADER from trantor.

I don't quite know what the `CMake templates` you mentioned is. Can you give me some links to documents or examples? I think the same problem in drogon needs to be solved.

CMakeLists.txt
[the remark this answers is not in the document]
OvermindDL1: > When I used find_package(OpenSSL CONFIG), I got an error on both MacOS and Linux:

Oh, that's an old style FindCMake file...  o.O

In that case it requires four lines instead of one for optional inclusion.  ^.^

```cmake
find_package(OpenSSL REQUIRED)
if(OpenSSL_FOUND)
    target_link_libraries(whatever_the_target_name_is PUBLIC OpenSSL::SSL OpenSSL::Crypto)
endif()
```
The `target_link_libraries` command will auto-link in the public headers, libraries, etc...  No need for the multiple manual commands.  :-)

> And if I remove the USE_OPENSSL macro, How do I handle conditional compilation like this？

A `USE_OPENSSL` macro would be appropriate here, but as it changes the API of the library this is one that you will want to bake into the INSTALL scripts as well via a template config.h.in or so.

CMakeLists.txt
OvermindDL1: CMake templates are very much like Autoconf templates (if you've used the old Makefiles build ecosystem), but instead of calling `AC_CONFIG_HEADERS` like in Autoconf, you instead call [`configure_file`](https://cmake.org/cmake/help/v3.15/command/configure_file.html) (link to the docs).

In essence, you will make a Template `config.h.in` file, with content like:
```cpp
#cmakedefine OpenSSL_FOUND
#cmakedefine OpenSSL_VERSION "@VAR2@"
#cmakedefine01 SOME_FLAG
```
If you do something like this in cmake:
```cmake
if(OpenSSL_FOUND)
  set(OpenSSL_VERSION "Blah")
endif()
option(SOME_FLAG "Set some flag" ON)
configure_file(config.h.in config.h @ONLY)
include_directories(${CMAKE_CURRENT_BINARY_DIR})
```
Will generate something like this:
```cpp
#define OpenSSL_FOUND
#define OpenSSL_VERSION "Blah"
#define SOME_FLAG 1
```
And if OpenSSL_FOUND is false and some_flag is off instead of on, then it would generate this:
```cpp
/* #undef OpenSSL_FOUND */
/* #undef OpenSSL_VERSION */
#define SOME_FLAG 0
```
It has a fairly simple set of primitive instructions (see docs) but it is actually quit powerful and perfect for all config-file needs, not just `#define`'ing as above but variable expansion and replacement and code-gen and more.  :-)

The purpose of it is to keep source code out of your build files.  You should never have source code in build files.  :-)

CMakeLists.txt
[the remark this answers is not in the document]
an-tao: I removed USE_OPENSSL macro from API header files now, this problem may be resolved.

trantor/utils/LogStream.h
OvermindDL1: > How about changing the line to #ifdef USE_STD_STRING_VIEW(If users want to enable the API, they should use cmake like cmake .. -DUSE_STD_STRING_VIEW)?

That would be far more common yeah.  :-)

> If the user library is compiled with C++14, they must use cmake .. -DCMAKE_CXX_STANDARD=14 to compile the trantor.

If you don't want to mandate a minimum C++ version that is fine, but if your library actually does require C++14 features then it is best to mandate that as a minimum regardless.

> I seem to understand some of your meaning, the trantor must now be distributed in the source code (compiled and installed by the user), if I modify it according to your principles, trantor seems to be available in the form of a library in addition to the source distribution (Header and library files), is my understanding correct?

In a fully and properly setup CMake build system, it would be possible for all of these to be done by the user:

* Grab a copy of the repo, compile and install it (which installs a perfect TrantorConfig.cmake system configuration file so no FindTrantor.cmake is needed, then they can use their globally installed version.
* Grab a copy of the repo, the user `add_subdirectory`'s it into their program and they can compile and use it all internally.
* You can distribute an 'installation' set of the library, which includes a set of libraries, header files, a FindTrantor.cmake file, etc...).
* The above installation set can be distributed as a tar.gz/zip or other compressed format, a deb file, a windows installer file, an rpm file, a whole variety of other things (all natively supported as install generators in CMake).

In general the first 2 points are generally sufficient, but once those are set up properly then adding in an installable distributable is fairly simple.  :-)

> Can I use the __cplusplus macro?

The `__cplusplus` macro used to be... not standard, but as of C++14 it is now a standard macro (C++14 is defined in it as `201402L` for example, C++17 is `201703L`).  Because it is a yearmonthqualifier it's not always fun to use though, so always double-check the values you check in it!  If you mandate a C++14 minimum version then it is always safe to use.  If you want to work on older C++'s then it 'might' work (on most compilers it will, though a variety of MSVC's are broken) but always check specific values so it at least falls back cleanly on compilers that don't support it.  I personally would use it to do conditional compilation based on standards version.

CMakeLists.txt
OvermindDL1: For note, you still don't need to make a config file just for something like `USE_OPENSSL`, the more usual method is to, say you have an SSLContext file in it, the entire API is exposed in the header, the `cpp` file for it contains the implementation, the general way would be to use a different cpp implementation file for if OpenSSL_FOUND compared to if it's not (where the SSLContext would be shimmed out).  This keeps the API identical regardless of how it is compiled (you of course want to throw an error if they try to turn on SSL without it existing in the shim).

an-tao: OvermindDL1 Yes, I see now. I just removed the USE_OPENSSL macro out of the public API header files.

an-tao: @OvermindDL1 And I will take some time to learn the CMake templates. Thank you for your explanation.

trantor/utils/LogStream.h
an-tao: @OvermindDL1 I can't agree more. I learned a lot. Thank you so much.
I mandate a C++14 minimum version, because I move some objects into lambdas(this is supported since C++14) in trantor.

an-tao:
@OvermindDL1 Do you think this PR can be merged? Thanks.

OvermindDL1:
Let me do another look-over, one sec...

CMakeLists.txt
```diff
@@ -1,27 +1,16 @@
...
-        MESSAGE(STATUS "c++17")
-    endif()
-endif()
+IF(NOT DEFINED CMAKE_CXX_STANDARD)
+    SET(CMAKE_CXX_STANDARD 14)
+ELSE()
+    IF(CMAKE_CXX_STANDARD LESS 14)
+        MESSAGE(FATAL_ERROR "C++14 is required")
+    ENDIF()
+ENDIF()
```
OvermindDL1: Lines 4 to 10 are still manually checking a global C++ version, I'd remove all these lines and down near where the project itself sets up its sources and links and so forth I'd add:
```cmake
set_property(TARGET trantor PROPERTY CXX_STANDARD 14)
```
This will mandate C++14 for your project, even if a user project requires higher.

In addition you could always set a default for the global `CMAKE_CXX_STANDARD` if it doesn't exist, but I prefer to leave that up to the users as a library should not be overriding their choices.

CMakeLists.txt
```diff
@@ -33,13 +22,9 @@ INCLUDE_DIRECTORIES(
...
-if(CMAKE_BUILD_TYPE STREQUAL "")
-    set(CMAKE_BUILD_TYPE Release)
-endif()
-
-set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Wall -std=${CMAKE_CXX_STD_FLAGS}")
-set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -Wall -std=${CMAKE_CXX_STD_FLAGS}")
-
+IF(CMAKE_BUILD_TYPE STREQUAL "")
+    SET(CMAKE_BUILD_TYPE Release)
+ENDIF()
```
OvermindDL1: This already has a pre-baked default of `Debug` and it should not be changed even if empty.  This setting should be controlled solely on the build command and not via a cmake file, thus these lines 25-27 should be removed.

CMakeLists.txt
```diff
@@ -48,75 +33,87 @@ LINK_DIRECTORIES(
...
+    trantor/net/inner/TimerQueue.cc)
+
+FIND_PACKAGE(OpenSSL)
+IF(OpenSSL_FOUND)
+    add_definitions(-DUSE_OPENSSL)
     include_directories(${OPENSSL_INCLUDE_DIR})
     link_libraries(${OPENSSL_LIBRARIES})
-    AUX_SOURCE_DIRECTORY(trantor/net/ssl DIR_SRC)
-endif()
+    SET(trantor_sources ${trantor_sources} trantor/net/ssl/SSLConnection.cc)
```
OvermindDL1: Fantastic!

CMakeLists.txt
```diff
@@ -48,75 +33,87 @@ LINK_DIRECTORIES(
...
-file(APPEND "${CONFIG_HEADER}" "using std::experimental::any_cast;\n\n")
-file(APPEND "${CONFIG_HEADER}" "#include <experimental/string_view>\n")
-file(APPEND "${CONFIG_HEADER}" "typedef std::experimental::basic_string_view<char> string_view;\n\n")
-
-endif()
-
-EXEC_PROGRAM(${PROJECT_SOURCE_DIR}/update_config.sh ARGS "${CONFIG_HEADER} ${PROJECT_SOURCE_DIR}/trantor/utils/config.h")
-
-if (MAKETEST STREQUAL YES)
+IF (MAKETEST STREQUAL YES)
```
OvermindDL1: This is a highly non-standard variable name for testing, is it something manually made?

If you want to create a proper `test` command for cmake then you should use the `ctest` functionality built into `cmake`.  It looks like all this does is build some programs that I'm guessing are manually run?  This is a good start for ctest, but ctest adds build commands for testing, timing reports, success runs, etc...  In essence if you setup a ctest setup in here then you will be able to run it just via `ctest --build-and-test . ./build_test --build-generator Ninja` (or whatever generator you want to use, I like ninja for its speed, it's so much faster than makefiles, though if you wanted makefiles then instead of `Ninja` you'd do `'Unix Makefiles'` instead, or whatever generator you want), or if it is just a normal build then you can just do `make test` and it will build and run the tests.

The basics to setup ctest in this file would be:
```cmake
enable_testing()
add_test(testname1 exename1 arg1 arg2 ...)
add_test(testname2 exename2 arg1 arg2 ...)
```
If you want to report results to a central cdash server you can replace `enable_testing()` with `include(CTest)` to set that up, but meh on that.

To actually build the programs that correspond to `exename1`/`exename2`/etc you just test if `BUILD_TESTING` is set, and if so then you build them like normal.  This variable is automatically defined if testing is requested.

Official `add_test` docs:  https://cmake.org/cmake/help/v3.15/command/add_test.html
`ctest` command docs:  https://cmake.org/cmake/help/latest/manual/ctest.1.html
I highly recommend this tutorial:  https://cgold.readthedocs.io/en/latest/tutorials/tests.html

Some other potentially relevant links from a quick google search:

* https://gitlab.kitware.com/cmake/community/wikis/doc/ctest/Testing-With-CTest
* http://bastian.rieck.me/blog/posts/2017/simple_unit_tests/
* https://a4z.bitbucket.io/blog/2018/05/17/Speed-up-your-test-cycles-with-CMake.html

CMakeLists.txt
```diff
@@ -48,75 +33,87 @@ LINK_DIRECTORIES(
...
+    trantor/net/inner/Connector.cc
+    trantor/net/inner/Poller.cc
+    trantor/net/inner/Socket.cc
+    trantor/net/inner/TcpConnectionImpl.cc
+    trantor/net/inner/Timer.cc
+    trantor/net/inner/TimerQueue.cc)
+
+FIND_PACKAGE(OpenSSL)
+IF(OpenSSL_FOUND)
+    add_definitions(-DUSE_OPENSSL)
```
OvermindDL1: Although this is entirely fine, but you could conditionally include a `SSLConnectionSkipped.cc` file instead, that exposes the same API but with no-ops (or errors) if called.  This is really what you want to do too if the API is exposed to users as you don't want the API changing depending on compiled-in features (otherwise 'that' should be another side library to optionally link in instead).

trantor/net/TcpConnection.h
```diff
@@ -58,20 +56,49 @@ class TcpConnection
...
+    void setContext(T &&context)
```
OvermindDL1: On a more programmatic and less build-system note, I question the moving into this function then forward constructing it style.  That's mandating an API that should not necessarily be necessary and will fail with many user types.  I would instead opt for full inline construction with perfect forwarding if you want to construct it inline, or better yet, just take a `std::shared_ptr` and make the user themselves construct it however they want (this wouldn't even need to be a template then).

trantor/net/TcpConnection.h
```diff
@@ -58,20 +56,49 @@ class TcpConnection
...
+    {
+        typedef typename std::remove_cv<
+            typename std::remove_reference<T>::type>::type ContextType;
+        _contextPtr = std::make_shared<ContextType>(std::forward<T>(context));
+    }
+    template <int N>
+    void setContext(const char (&str)[N])
```
OvermindDL1: On a more programmatic and less build-system note, although this is obviously to make taking a statically-sized string a little easier, it seems a bit special-case'y, as above I'd opt to just accept a `std::shared_ptr` of any internal type instead and let the user construct it as they wish.

trantor/net/TcpConnection.h
```diff
@@ -58,20 +56,49 @@ class TcpConnection
...
+    {
+        _contextPtr = std::make_shared<const char *>(str);
+    }
+    template <typename T>
+    const T &getContext() const
```
OvermindDL1: This does make the user API a bit easier, but an assert is always a sign of a user bug waiting to happen.  Instead of returning a reference to the object (especially as the object may not exist), it should just return the pointer to it directly or `nullptr`, though I'd personally opt for returning a new invocation of the shared_ptr itself (with increased reference count) via a function named `getContextRC` and name the one that returns a potentially null pointer as `getContextOpt` or so.  Personally I utterly hate loose pointers in public API's so would instead opt for `std::optional` if this were C++17, but C++14 is the minimum required here so a loose pointer would do.  I would absolutely not return a reference though as that is an API that implies that it can never be null (and in fact if it is called at release without asserts the compiler is free to optimize the code in a way that it can assume it will not be null, which can make the shared_ptr checks and all just vanish in many cases, which is highly unsafe).

trantor/net/TcpConnection.h
```diff
@@ -58,20 +56,49 @@ class TcpConnection
...
-
+    template <typename T>
+    T &getContext()
+    {
+        assert(_contextPtr);
+        return (*(T *)_contextPtr.get());
+    }
+    bool hasContext()
+    {
+        return (bool)_contextPtr;
```
OvermindDL1: Casting to bool even via `operator bool` is not always a safe action as a shared_ptr could potentially own `null` memory, thus returning `true` even when the internal pointer is `nullptr`.  Especially if you have a potential reference access but in any case you want to both test if the shared_ptr is empty and also if its internal pointer is not `nullptr`.

CMakeLists.txt
an-tao: I put several unit tests(with **gtest** lib) into the test branch. And these tests in master branch are manually run as you guessed. 
I don't know much about test tools. Can CTest run the client and the server test at same time  automatically? 
Normalizing these tests seems like a big project. Can I do them in the next PR? ^.^

trantor/net/TcpConnection.h
an-tao: The API can accept copyable or movable type, if user want to store some un-copyable and un-movable objects in the context(such as std::mutex), he should swap it with smart pointer  and then use `setContext`.
Compared with a `setContext(const std::shared_ptr<void>&)` interface, this one has less annoying type conversions in use (that is done in the getContext function).
As far as type compatibility is concerned, the current interface can accept copyable left-value references or movable right-value references, bare pointers, smart pointers or string literals. I think it can be adapted to most scenarios.

trantor/net/TcpConnection.h
an-tao: @OvermindDL1 I've modified the methods.
For convenience, I keep the template form of the getter

CMakeLists.txt
an-tao: @OvermindDL1  I removed the lines and use `set_property(TARGET trantor PROPERTY CXX_STANDARD 14)`, but for test targets, I have to add a long list to the set_property command, Is there an easier way to do this?
https://github.com/an-tao/trantor/blob/2eb8efb56543a5bbaf0ae8bee6f907c19cf86a7e/trantor/tests/CMakeLists.txt#L38-L55

CMakeLists.txt
an-tao: I removed them.

CMakeLists.txt
an-tao: I added a `SSLConnectionSkipped.cc` file and removed the `USE_OPENSSL`

an-tao:
@OvermindDL1 Please check it again. I believe it's much better now ^.^
Thanks!

CMakeLists.txt
```diff
@@ -33,90 +11,98 @@ INCLUDE_DIRECTORIES(
...
-    link_libraries(${OPENSSL_LIBRARIES})
-    AUX_SOURCE_DIRECTORY(trantor/net/ssl DIR_SRC)
-endif()
+    LINK_LIBRARIES(${OPENSSL_LIBRARIES})
+    SET(trantor_sources ${trantor_sources} trantor/net/ssl/SSLConnection.cc)
+ELSE()
+    SET(trantor_sources ${trantor_sources} trantor/net/ssl/SSLConnectionSkipped.cc)
+ENDIF()
 
 IF (CMAKE_SYSTEM_NAME MATCHES "Linux")
```
OvermindDL1: I'm wondering if this would be better to do in-source, to include the right file depending on the OS, instead of having cmake handle it.  As it stands you won't be able to cross-compile to linux from something else, if that is a consideration you want to worry about...

trantor/net/ssl/SSLConnectionSkipped.cc
```diff
@@ -0,0 +1,80 @@
...
+class SSLContext
+{
+};
+class SSLConn
+{
+};
+
+void initOpenSSL()
+{
+    LOG_FATAL << "SSL is not supported!";
```
OvermindDL1: Hmm, I'd probably word these 3 log strings as something like `"SSL is not supported due to not compiling with OpenSSL"` or so, to give a hint of what they need to do to fix it.

CMakeLists.txt
an-tao: Yes, you are right, I will find a better way. Thanks!

an-tao:
@OvermindDL1 Thank you very much.  It looks a lot better with your help.

PR #45, opened 2019-08-05
Add the TrantorConfig.cmake

an-tao:


an-tao:
@OvermindDL1 I have added the TrantorConfig.cmake file, but I am not sure if it is completely correct, please help me check it. Thanks a lot.

an-tao:
reopened at #46

PR #46, opened 2019-08-05
Add the TrantorConfig.cmake

an-tao:


an-tao:
@OvermindDL1 I have added the TrantorConfig.cmake file, but I am not sure if it is completely correct, please help me check it. Thanks a lot.

CMakeLists.txt
```diff
@@ -89,7 +84,72 @@ IF (MAKETEST STREQUAL YES)
...
+# Export the package for use from the build-tree
+# (this registers the build-tree with a global CMake-registry)
+export(PACKAGE Trantor)
+
+# Create the TrantorConfig.cmake and TrantorConfigVersion files
+file(RELATIVE_PATH REL_INCLUDE_DIR "${INSTALL_TRANTOR_CMAKE_DIR}"
+   "${INSTALL_INCLUDE_DIR}")
+# ... for the build tree
+set(CONF_INCLUDE_DIRS "${PROJECT_SOURCE_DIR}" "${PROJECT_BINARY_DIR}")
+configure_file(cmake/templates/TrantorConfig.cmake.in
```
OvermindDL1: You don't need this, or the `TrantorConfig.cmake.in` file, CMake can generate it fully.  Above on the `export` line it can autogenerate it.  The basic format of those is:
```
export(TARGETS
        your_target
    NAMESPACE YourPackage::
    FILE "${CMAKE_CURRENT_BINARY_DIR}/YourPackageConfig.cmake"
)
install(EXPORT
        YourPackageConfig
    DESTINATION "${CMAKE_INSTALL_DATADIR}/YourPackage/cmake"
    NAMESPACE YourPackage::
)
```

OvermindDL1: You don't need the namespace if you don't have components, but I find it good form to always use it anyway in case components are added later.

an-tao: @OvermindDL1 , I will try it tomorrow. Thank you very much!

an-tao:
@OvermindDL1 I tried it today with the method you introduced. It doesn't seem to work, I can't figure out what went wrong :-(
I put them in the `cmake_config_test` branch.

OvermindDL1:
> I tried it today with the method you introduced. It doesn't seem to work, I can't figure out what went wrong :-(

Oy, don't know what I was thinking yesterday...  o.O

Yes you do need the *config.cmake.in file, but instead of `configure_file` on line 135 you need to use `configure_package_config_file`, see docs at:  https://cmake.org/cmake/help/v3.15/module/CMakePackageConfigHelpers.html
Using `write_basic_package_version_file` to make a corresponding version file is highly recommended too (same docs).

See example at the bottom of the docs.  >.>

an-tao:
I have another problem. In drogon, I added trantor as a submodule, so in drogon's CMakeLists, I added the following lines:
```
ADD_SUBDIRECTORY(trantor)
set(Trantor_DIR ${PROJECT_BINARY_DIR}/trantor)
FIND_PACKAGE(Trantor CONFIG REQUIRED NO_DEFAULT_PATH)
include_directories(${TRANTOR_INCLUDE_DIRS})
link_libraries(${TRANTOR_LIBRARIES})
```
This works, but it's not so convenient. How can I just add a line of `ADD_SUBDIRECTORY(trantor)` to make everything work?

OvermindDL1:
If it's locally included, like via git submodules, then yes, `add_subdirectory` is the recommended way.  The above set of commands is only recommended for externally built packages.

an-tao:
@OvermindDL1 After testing, I think it can work now.

PR #63, opened 2020-01-14
Alfa port to Windows of Trantor

mcirsta:
Some things are probably wrong here, especially in the epoll I think but this is not really my area of expertise.


an-tao:
OK！I'll try to fix it. thank you very much！

mcirsta:
> OK！I'll try to fix it. thank you very much！

Thanks. I am also trying to figure out how it works but it will take me some time to get familiar with the code.
 If you have any questions let me know.

mcirsta:
One thing that maybe obvious but it wasn't to me is that in msys2 you have to install the package versions that have "mingw" in the name. There are also some internal ones or something.

an-tao:
I installed a msys2, but I found that cmake does not work in this environment (there is no output after running `cmake ..`). Did I miss anything?
In addition, can we consider compiling with visual studio C++(which now supports cmake project) ?

mcirsta:
I will provide full instructions for this. Sorry, should have done it from the beginning.

 I tried VS 2019 too but there's just too many things missing. Mostly it's the same things that need to work so I think the best plan is this. Get this working with msys2 because it's the easiest way to get a Windows version that works and then do even further modifications to get a VS 2019 port working.

an-tao:
@mcirsta ,I made a new branch named win_vs, which can be successfully compiled with VS2019 (but something went wrong with EventLoop), please check if it helps you.

an-tao:
@mcirsta I've invited you to join trantor's collaborators.

mcirsta:
Thanks, I accepted the invitation. Cool , I see you made it compile with VS 2019. For some reason I thought it was harder, I think I have more initial issues with this.
 One issue I remember having is the needed libraries, OpenSSL and jsconcpp. How did you deal with these ? Should they be added to third party and compiled ?

an-tao:
@mcirsta , jsoncpp is not the dependency of trantor. and we could ignore the OpenSSL first and make the normal function of trantor works.
We can use [vcpkg] (https://github.com/microsoft/vcpkg/blob/master/docs/index.md) to manage dependencies, and although I haven't used it, it looks good.

mcirsta:
Meanwhile just in case you need them here are the instructions for msys2 ( I suggest starting with a clean msys2 ). They're written for users so include every step.  I checked that both drogon and trantor work with these.

1. Install msys2 64 bit from https://www.msys2.org/
2. Open the msys2 64 bit shell and do a "pacman -Suy" to update . After it's done you will have to close the shell and restart it
3. Install git, devel tools and needed libraries: "pacman -S git mingw-w64-x86_64-cmake mingw-w64-x86_64-gcc mingw-w64-x86_64-make mingw-w64-x86_64-sqlite3 mingw-w64-x86_64-postgresql mingw-w64-x86_64-dlfcn"
4. git clone --recursive https://github.com/mcirsta/drogon.git
5. Go to drogon folder and make a "build" folder
6. Go to build folder and call cmake with: "/mingw64/bin/cmake.exe -G "MSYS Makefiles" -DCMAKE_MAKE_PROGRAM=mingw32-make  -DBUILD_EXAMPLES=ON -DBUILD_TESTING=ON -DCMAKE_BUILD_TYPE=Debug .."
7. Build with: "/mingw64/bin/mingw32-make.exe -j4"

mcirsta:
Also it now needs : "mingw-w64-x86_64-gtest"

an-tao:
Thanks, now I can build the win_vs branch with msys2 without compilation error. But the tcp server doesn't seem to work. I have to take some time to fix it.

PR #65, opened 2020-02-07
SSL_CTX_use_certificate_file() -> SSL_CTX_use_certificate_chain_file()

ihmc3jn09hk:
According to https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_use_certificate_file.html, it is preferred to use SSL_CTX_use_certificate_chain_file() instead.


an-tao:
@ihmc3jn09hk Thank your very much.

PR #81, opened 2020-04-16
Add fromDbStringLocal()

ihmc3jn09hk:
Helper function
Maybe operators like +, - shall be added later

### DIFF
--- a/trantor/utils/Funcs.h
+++ b/trantor/utils/Funcs.h
@@ -31,7 +31,7 @@ inline uint64_t ntoh64(uint64_t n)
 {
     return hton64(n);
 }
-inline std::vector<std::string> strsplit(const std::string &s, const std::string &delimiter )
+inline std::vector<std::string> splitString(const std::string &s, const std::string &delimiter )
 {
   size_t last = 0;
   size_t next = 0;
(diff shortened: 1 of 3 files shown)


trantor/utils/Funcs.h
```diff
@@ -30,4 +31,16 @@ inline uint64_t ntoh64(uint64_t n)
 {
     return hton64(n);
 }
+inline std::vector<std::string> strsplit(const std::string &s, const std::string &delimiter )
```
an-tao, line 34: this is implemented in drogon:
https://github.com/an-tao/drogon/blob/17b8c337f9fe6513e81fa3784bc7559cd7f8fcb7/lib/inc/drogon/utils/Utilities.h#L48-L51

ihmc3jn09hk: I see there is an implementation in drogon. However, trantor is not depended on drogon. Therefore, I reimplemented it and see drogon's splitString wrap on trantor implementation?
Or would it be good to move it to trantor?

an-tao: Oh, I see, this is ok, please use Hump nomenclature. (change strsplit to splitString)

ihmc3jn09hk: I will update that name and wrap the Drogon's functions if needed.

an-tao: Ok! thanks

PR #83, opened 2020-04-17
Fix a bug in the splitString() function

an-tao:


an-tao:
@ihmc3jn09hk please check this. We'd better add a unittest for this.

PR #91, opened 2020-05-27
Optimize the LockFreeQueue

an-tao:


an-tao:
@rbugajewski the lock-free multiple producers single consumer queue is the lowest component of the EventLoop, recently I found it can be improved in some way, I use pointer instead of object to avoid unuseful constructing, I'm not sure but I think it works. The complexity introduced by atomics or memory barries make me have to be careful. Please take a look on this PR. thanks.
Note: Only the FreeLockQueue.h file was changed, others were only formatted.

rbugajewski:
I hope I can take a look at it today, or tomorrow at latest.

an-tao:
@rbugajewski Thanks so much!

PR #94, opened 2020-06-09
Add support for Solaris

vedranmiletic:
Tested on illumos distribution OmniOS Community Edition.


an-tao:
@vedranmiletic Thanks so much.

PR #95, opened 2020-06-28
fixed install shared library errors.

linuxserverdev:
when BUILD_TRANTOR_SHARED on,cmake triggered an error


an-tao:
Thanks so much!

PR #105, opened 2020-12-12
Fix MinGW error with inet_ntop and inet_pton

ivanka2012:
Header error only, this shouldn't ruin linking with Drogon or with any other library


an-tao:
@ivanka2012 After I update trantor to the latest version in Drogon, the [appveyor](https://ci.appveyor.com/project/an-tao/drogon/builds/36808571/job/rhucbfg1rlutfes6) CI reported errors when building. Please check it, thanks.

PR #106, opened 2020-12-13
Fix a macro regression when using MSVC

ivanka2012:
A bit of an embarrassing one :grimacing:. Let's hope no one has used the current version in the meantime

https://ci.appveyor.com/project/an-tao/drogon/builds/36809149


an-tao:
> A bit of an embarrassing one 😬. Let's hope no one has used the current version in the meantime
> 
> https://ci.appveyor.com/project/an-tao/drogon/builds/36809149

Don't worry, the last MinGW patch has not been used in drogon.

PR #110, opened 2020-12-23
Add support of VS2019

BertrandD:
The objective is to be able to compile with newer version of Visual Studio MSVC (currently only working with Visual Studio 2015 (14.0))


an-tao:
@BertrandD Thanks so much for your patch.

PR #117, opened 2021-01-19
add secondsSinceEpoch to trantor::Date

### LINKED ISSUE #114: [FR] 32bit timestamp conversion
We already have `trantor::Date::microSecondsSinceEpoch` but it will be a pain in ass if I want seconds instead of microseconds. 
It will be great if you can implement such function like `trantor::Date::secondsSinceEpoch`.

kotori2:
Fix #114

### DIFF
--- a/trantor/utils/Date.h
+++ b/trantor/utils/Date.h
@@ -168,6 +168,16 @@ class Date
     }
 
     /**
+     * @brief Get the number of seconds since 1970-01-01 00:00.
+     *
+     * @return int64_t
+     */
+    int64_t secondsSinceEpoch() const
+    {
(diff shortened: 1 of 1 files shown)


trantor/utils/Date.h
```diff
@@ -166,6 +166,16 @@ class Date
     {
         return microSecondsSinceEpoch_;
     }
+  
+    /**
+     * @brief Get the number of seconds since 1970-01-01 00:00.
+     *
+     * @return uint32_t
+     */
+    uint32_t secondsSinceEpoch() const
```
an-tao, line 175: Please use the same type of microSecondsSinceEpoch_ to avoid type conversion.

an-tao (changes requested):

an-tao (approved):

Issue #115, opened 2021-01-18
Is there a way to include the SNI to the ssl?


tolysz:
I was wondering what could be done to support virtual hosts.
https://www.openssl.org/docs/man1.1.1/man3/SSL_get_servername_type.html

PR #120, opened 2021-02-08
SSL certificate validation

marty1885:
Sorry for sending a patch just before CNY.

This patch enables trantor to validate the peer's SSL certificate. Drogon need additional patch to make use of this patch. I'm not sure if this is how it should be done. Feel free to comment and I'll update the code.

**Limitation**

It does not validate the host name (yet) as that would require serious rework of the core. But still better than nothing.

Also, I can't find a way to trigger the error callback after detecting a cert error. Need help here.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -50,6 +50,7 @@ jobs:
         sudo apt update
         # These aren't available or don't work well in vcpkg
         sudo apt install openssl libssl-dev
+        sudo apt install dos2unix
 
     - name: install gtest
       run: |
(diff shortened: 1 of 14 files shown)


marty1885:
The drogon part is available at https://github.com/marty1885/drogon **validate-cert** branch. I haven't make a PR yet because CI is gonna fail without updated trantor.

an-tao:
@marty1885 Thanks so much for the patch, would you like to take a look at the issue #115, maybe it could be resolved together?

marty1885:
I don't think so. At least not without rearchicturing both trantor and drogon. (At minimum. Drogon runs off a single global instance. Whch makes virtual hosts pointless)

May you look into triggering the error cb? I don't know how to do it.

PR #123, opened 2021-02-23
fix string concat bug

marty1885:
Fixes a string concatenation bug


marty1885:
@an-tao Please review

PR #127, opened 2021-03-12
Add exports macro to allow Shared Library with hidden symbols by default (e.g. Windows)

BertrandD:
If you build drogon as a shared library with `CMAKE_CXX_VISIBILITY_PRESET` set to `hidden` and `CMAKE_VISIBILITY_INLINES_HIDDEN` to `1` all the symbols are hidden, thus linking a project with the drogon shared library will fail.

On Windows, the symbols are hidden by default and you need to export them explicitely. 

This PR enable building trantor as a shared library on Windows.

An abort() is raised when closing the project, I need to check if it is related to my work or not. In the meantime, I am interested in your feedback.


an-tao:
@BertrandD Thanks for the PR, I'm not sure but do we need to install the `exports.h` to the trantor install path?

BertrandD:
> @BertrandD Thanks for the PR, I'm not sure but do we need to install the `exports.h` to the trantor install path?

Sure ! Fixed

trantor/utils/AsyncFileLogger.h
```diff
@@ -43,19 +44,19 @@ class AsyncFileLogger : NonCopyable
      * @param msg
      * @param len
      */
-    void output(const char *msg, const uint64_t len);
+    void TRANTOR_EXPORT output(const char *msg, const uint64_t len);
```
an-tao, line 47: Please make the location of TRANTOR_EXPORT to be consistent (e.g. place them all before the return type of functions)

trantor/utils/SerialTaskQueue.h
```diff
@@ -61,7 +62,7 @@ class SerialTaskQueue : public TaskQueue
      *
      * @param name
      */
-    explicit SerialTaskQueue(const std::string &name = std::string());
+    explicit SerialTaskQueue(const std::string &name);
```
an-tao, line 65: Why remove the default value here?

BertrandD: When you cann `SerialTaskQueue();` there is an ambiguity between these two constructors:

```
    SerialTaskQueue() = delete;
``` 
and 
```
    explicit SerialTaskQueue(const std::string &name = std::string());
```

Both constructors can be called with no parameters.

an-tao: I see, you are right. Thanks.

an-tao:
@OvermindDL1 Hi~ Would you like to take a look at this PR? Thank you very much!

trantor/tests/CMakeLists.txt
```diff
@@ -44,6 +44,8 @@ set_property(TARGET ${targets_list} PROPERTY CXX_STANDARD 14)
 set_property(TARGET ${targets_list} PROPERTY CXX_STANDARD_REQUIRED ON)
 set_property(TARGET ${targets_list} PROPERTY CXX_EXTENSIONS OFF)
 
+find_package(Threads)
+
 foreach(T ${targets_list})
-  target_link_libraries(${T} PRIVATE trantor)
+  target_link_libraries(${T} PRIVATE trantor Threads::Threads)
```
an-tao, line 50: There are some lines in the main CMakeLists:
```
if(WIN32)
  target_link_libraries(${PROJECT_NAME} PRIVATE ws2_32 Rpcrt4)
  if(OpenSSL_FOUND)
    target_link_libraries(${PROJECT_NAME} PRIVATE Crypt32 Secur32)
  endif(OpenSSL_FOUND)
else(WIN32)
  target_link_libraries(${PROJECT_NAME} PRIVATE pthread)
endif(WIN32)
``` 
should we put this to private libraries of trantor?

an-tao: @BertrandD How do you think about this?

BertrandD: Uh, I forgot to reply. Sorry.

You're right, I'll add Threads::Threads as private dep of trantor instead of pthread

an-tao:
@BertrandD Merged! Thanks for this pr, please update the related PR of Drogon.

PR #128, opened 2021-03-18
add isUnspecified() to indicate if parsing failed

marty1885:
Throws an exception when `InetAddress::InetAddress` failed to initialize. I'll make a PR on Drogon when this is merged. 

Refer to https://github.com/an-tao/drogon/issues/755 for detail

### DIFF
--- a/trantor/unittests/InetAddressUnittest.cc
+++ b/trantor/unittests/InetAddressUnittest.cc
@@ -11,7 +11,8 @@ TEST(InetAddress, innerIpTest)
     EXPECT_EQ(true, InetAddress("10.0.0.1", 0).isIntranetIp());
     EXPECT_EQ(true, InetAddress("172.31.10.1", 0).isIntranetIp());
     EXPECT_EQ(true, InetAddress("127.0.0.1", 0).isIntranetIp());
-    EXPECT_EQ(false, InetAddress("example.com", 0).isUnspecified());
+    EXPECT_EQ(true, InetAddress("example.com", 0).isUnspecified());
+    EXPECT_EQ(false, InetAddress("127.0.0.2", 0).isUnspecified());
 }
 int main(int argc, char **argv)
 {
--- a/trantor/unittests/InetAddressUnittest.cc
+++ b/trantor/unittests/InetAddressUnittest.cc
@@ -13,6 +13,7 @@ TEST(InetAddress, innerIpTest)
     EXPECT_EQ(true, InetAddress("127.0.0.1", 0).isIntranetIp());
     EXPECT_EQ(true, InetAddress("example.com", 0).isUnspecified());
     EXPECT_EQ(false, InetAddress("127.0.0.2", 0).isUnspecified());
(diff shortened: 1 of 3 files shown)


trantor/unittests/InetAddressUnittest.cc
```diff
@@ -11,6 +11,7 @@ TEST(InetAddress, innerIpTest)
     EXPECT_EQ(true, InetAddress("10.0.0.1", 0).isIntranetIp());
     EXPECT_EQ(true, InetAddress("172.31.10.1", 0).isIntranetIp());
     EXPECT_EQ(true, InetAddress("127.0.0.1", 0).isIntranetIp());
+    EXPECT_EQ(false, InetAddress("example.com", 0).isUnspecified());
```
an-tao, line 14: This line should be true

an-tao (changes requested):

trantor/unittests/InetAddressUnittest.cc
an-tao: And please add a positive test case.

marty1885: This should be false since InetAddress can't resolve FQDN to an IP address.

an-tao:
I think the status should also be changed when other ctors and setters being called

marty1885:
Updated. My bad

trantor/unittests/InetAddressUnittest.cc
```diff
@@ -11,6 +11,9 @@ TEST(InetAddress, innerIpTest)
     EXPECT_EQ(true, InetAddress("10.0.0.1", 0).isIntranetIp());
     EXPECT_EQ(true, InetAddress("172.31.10.1", 0).isIntranetIp());
     EXPECT_EQ(true, InetAddress("127.0.0.1", 0).isIntranetIp());
+    EXPECT_EQ(true, InetAddress("example.com", 0).isUnspecified());
+    EXPECT_EQ(false, InetAddress("127.0.0.2", 0).isUnspecified());
```
hellow554, line 15: This is wrong. The range is `127.0.0.0/8`

marty1885: Unspecified means `InetAddress` failed to parse the supplied address. Akin to how ASIO works.

hellow554: IMHO it doens't matter what ASIO does.
[RFC 1122 Section 3.2.1.3 g](https://tools.ietf.org/html/rfc1122#section-3.2.1.3) defines what a loopback adress looks like so the RFC should be the reference not any Software implementation.

hellow554: I think Rust did a pretty good job ob describing and implemeting ipv4 and ipv6
https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html
https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html

PR #129, opened 2021-03-19
fix: Expired timer' id not reclaimed

ALLReadyGo:
@file: TimerQueue.cc


ALLReadyGo (commented):
my mistake
Logger.cc Should be modified like this
`
uint64_t now = date_.secondsSinceEpoch();
uint64_t microSec = 
        date_.microSecondsSinceEpoch() - date_.roundSecond().microSecondsSinceEpoch();
`

PR #130, opened 2021-03-19
fix: Timer'id is not recycled in timerQueue

ALLReadyGo:
If the timer has been called and not repeat, it should be removed from the TimerQueue.timerIdSet_

### DIFF
--- a/trantor/net/inner/TimerQueue.cc
+++ b/trantor/net/inner/TimerQueue.cc
@@ -270,13 +270,8 @@ void TimerQueue::reset(const std::vector<TimerPtr> &expired,
     loop_->assertInLoopThread();
     for (auto const &timerPtr : expired)
     {
-        // if (timerPtr->isRepeat() &&
-        //     timerIdSet_.find(timerPtr->id()) != timerIdSet_.end())
-        // {
-        //     timerPtr->restart(now);
-        //     insert(timerPtr);
-        // }
-        if (timerIdSet_.find(timerPtr->id()) != timerIdSet_.end())
+        auto iter = timerIdSet_.find(timerPtr->id());
+        if (iter != timerIdSet_.end())
         {
             if (timerPtr->isRepeat())
             {
@@ -285,9 +280,7 @@ void TimerQueue::reset(const std::vector<TimerPtr> &expired,
             }
             else
             {
-                // If the timer has been called and not repeat, it should be removed from the set
-                // I guess you forget that 
-                timerIdSet_.erase(timerPtr->id());
+                timerIdSet_.erase(iter);
             }
         }
         
--- a/trantor/net/inner/TimerQueue.cc
+++ b/trantor/net/inner/TimerQueue.cc
@@ -27,7 +27,7 @@
 
 using namespace trantor;
 #ifdef __linux__
-int createTimerfd()
+static int createTimerfd()
 {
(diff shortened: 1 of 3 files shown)


trantor/net/inner/TimerQueue.cc
```diff
@@ -270,12 +270,27 @@ void TimerQueue::reset(const std::vector<TimerPtr> &expired,
...
     for (auto const &timerPtr : expired)
     {
-        if (timerPtr->isRepeat() &&
-            timerIdSet_.find(timerPtr->id()) != timerIdSet_.end())
+        // if (timerPtr->isRepeat() &&
+        //     timerIdSet_.find(timerPtr->id()) != timerIdSet_.end())
+        // {
+        //     timerPtr->restart(now);
+        //     insert(timerPtr);
+        // }
```
an-tao, line 278: Please delete useless lines and don't worry, because we have git version control.

trantor/net/inner/TimerQueue.cc
```diff
@@ -270,12 +270,27 @@ void TimerQueue::reset(const std::vector<TimerPtr> &expired,
...
-            insert(timerPtr);
+            if (timerPtr->isRepeat())
+            {
+                timerPtr->restart(now);
+                insert(timerPtr);
+            }
+            else
+            {
+                // If the timer has been called and not repeat, it should be removed from the set
+                // I guess you forget that
```
an-tao, line 289: Yes, you are right, thanks so much, would u please remove the `I guess ..` line, we can discuss in github instead of in comments of source code :)

trantor/net/inner/TimerQueue.cc
```diff
@@ -270,12 +270,27 @@ void TimerQueue::reset(const std::vector<TimerPtr> &expired,
...
     {
-        if (timerPtr->isRepeat() &&
-            timerIdSet_.find(timerPtr->id()) != timerIdSet_.end())
+        // if (timerPtr->isRepeat() &&
+        //     timerIdSet_.find(timerPtr->id()) != timerIdSet_.end())
+        // {
+        //     timerPtr->restart(now);
+        //     insert(timerPtr);
+        // }
+        if (timerIdSet_.find(timerPtr->id()) != timerIdSet_.end())
```
an-tao, line 279: Usually, I use an iterator for this, because it may be used later:
```c++
auto iter=timerIdSet_.find(timerPtr->id());
if(iter!=timerIdSet_.end())
{
      if().....
      else
      {
           timerIdSet_.erase(iter); //use an iterator to avoid a search same as before.
      }
}
```

trantor/utils/Logger.cc
```diff
@@ -76,9 +76,14 @@ static thread_local uint64_t threadId_{0};
 
 void Logger::formatTime()
 {
-    uint64_t now = date_.microSecondsSinceEpoch();
-    uint64_t microSec = now % 1000000;
-    now = now / 1000000;
+    // uint64_t now = date_.microSecondsSinceEpoch();
+    // uint64_t microSec = now % 1000000;
+    // now = now / 1000000;
+
```
an-tao, line 82: Please remove these lines.

an-tao:
@ ALLReadyGo Thanks so much for the patch.
BTW, you don't need to close this PR and reopen a new one, just make new commits on your branch.

PR #134, opened 2021-03-28
Update AsyncFileLogger.cc

wbvalid:
Since at line no.50 the threadPtr_ already joined, it means that tmpBuffers_ as well as writeBuffers_ are already empty. There's no need to push the current logBufferPtr_ back to writeBuffers_ and pop it out again.


an-tao:
@wbvalid Thanks for your patch, but IMO, the writeBuffers_ is not necessarily empty because other threads may call output function and swap new data into writeBuffers_(line no.80) after the writing thread being joined. This is the last chance to write them into the log file.
In the no.59 line 'while' should be used instead of 'if'.

PR #142, opened 2021-04-21
gtest_discover_tests requires cmake 3.10

pkovacs:


hks2002:
Yes. it shoud be upgrade versionto 3.10

PR #143, opened 2021-04-22
GNU: -Wall -Wextra -Werror; fix related warnings

pkovacs:
for GNU compiler, add `-Wall -Wextra -Werror` and fix all related warnings

### DIFF
--- a/trantor/net/EventLoop.cc
+++ b/trantor/net/EventLoop.cc
@@ -68,12 +68,11 @@ EventLoop::EventLoop()
       currentActiveChannel_(nullptr),
       eventHandling_(false),
       timerQueue_(new TimerQueue(this)),
-      threadLocalLoopPtr_(&t_loopInThisThread)
 #ifdef __linux__
-      ,
       wakeupFd_(createEventfd()),
-      wakeupChannelPtr_(new Channel(this, wakeupFd_))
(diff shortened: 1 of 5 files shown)


trantor/net/EventLoop.cc
```diff
@@ -304,11 +303,11 @@ void EventLoop::wakeup()
     //     return;
     uint64_t tmp = 1;
 #ifdef __linux__
-    int ret = write(wakeupFd_, &tmp, sizeof(tmp));
+    __attribute__((unused)) int ret = write(wakeupFd_, &tmp, sizeof(tmp));
```
an-tao, line 306: __attribute__((unused)) is a GCC extension, please use `(void)ret` instead of it to avoid compilation errors if someone uses other compilers in Linux.

an-tao:
@pkovacs Thanks so much.

trantor/net/EventLoop.cc
pkovacs: @an-tao done!   the compilers that come to mind for linux -- gnu, clang and icc all accept `__attribute((unused))__`, but the void cast is more universal.  there's also `[[maybe_unused]]` for c++17, but that involves even more checks, so void it is.

PR #148, opened 2021-05-26
Add SSL_CONF_cmd support

oToToT:
Add some code to check the design is fine.

This is intended to implement the feature described in an-tao/drogon#761.


an-tao:
@oToToT Thanks so much, merged!

PR #150, opened 2021-05-26
Allow OpenSSL recognize SSL_CONF_cmd options

oToToT:
OpenSSL requires either SSL_CONF_FLAG_CMDLINE, SSL_CONF_FLAG_FILE, or both be enabled to recognize SSL_CONF_cmd options.

However, I didn't notice that in an-tao/trantor#148, so here is a fix for that.

Sorry for not checking it carefully then.

See https://www.openssl.org/docs/manmaster/man3/SSL_CONF_CTX_set_flags.html and 
https://www.openssl.org/docs/manmaster/man3/SSL_CONF_cmd.html for more detail.

### DIFF
--- a/trantor/net/inner/TcpConnectionImpl.cc
+++ b/trantor/net/inner/TcpConnectionImpl.cc
@@ -209,7 +209,6 @@ class SSLContext
         SSL_CONF_CTX_set_flags(cctx, SSL_CONF_FLAG_CLIENT);
         SSL_CONF_CTX_set_flags(cctx, SSL_CONF_FLAG_CERTIFICATE);
         SSL_CONF_CTX_set_flags(cctx, SSL_CONF_FLAG_FILE);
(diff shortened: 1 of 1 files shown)


oToToT:
Oops, seems like the behavior when both flags enabled on alpine3.13 is quite weird. I should have check.
Another way to avoid this is to only allow one kind of option name, but which should we allow?

an-tao:
> Oops, seems like the behavior when both flags enabled on alpine3.13 is quite weird. I should have check.
> Another way to avoid this is to only allow one kind of option name, but which should we allow?

I prefer the SSL_CONF_FLAG_FILE if we can't set both.

oToToT:
OK, it's not related to alpine3.13.
OpenSSL seems only allow one flag to be enabled now. (see openssl/openssl#15508).

Then, I would try only enabling `SSL_CONF_FLAG_FILE` and document it somewhere.

PR #152, opened 2021-06-10
added a try_compile block to detect if we need to link against atomic

jgilje:
This is required at least on linux and arm32 (like a raspberry pi). This issue is also mentioned in https://github.com/an-tao/drogon/issues/517.

The try_compile block was mostly just lifted from https://github.com/WebPlatformForEmbedded/WPEWebKitLauncher/blob/master/CMakeLists.txt


an-tao:
@jgilje Thanks so much.

PR #156, opened 2021-07-07
Added support for paths containing unicode characters on Windows

Greisby:
Converts std::string paths, considered as being utf-8, to std::wstring when creating/renaming log files, and when sending files (TcpConnection::sendFile).
Added a new entry TcpConnection::sendFile with wchar_t* on windows, for use in drogon.

PR #158, opened 2021-07-13
Feature/support windows unicode paths

Greisby:
[PR #156](https://github.com/an-tao/trantor/pull/156) continued.


Greisby:
Hi, notice that I made changes to CMakeList.txt to that headers are visible in Visual Studio & Xcode.

CMakeLists.txt
```diff
@@ -108,10 +109,32 @@ if(c-ares_FOUND)
   message(STATUS "c-ares found!")
   target_link_libraries(${PROJECT_NAME} PRIVATE c-ares_lib)
   set(TRANTOR_SOURCES ${TRANTOR_SOURCES} trantor/net/inner/AresResolver.cc)
-else()
+else(c-ares_FOUND)
   set(TRANTOR_SOURCES ${TRANTOR_SOURCES} trantor/net/inner/NormalResolver.cc)
+endif(c-ares_FOUND)
+
+if ((CMAKE_GENERATOR MATCHES "Visual Studio") OR (CMAKE_GENERATOR MATCHES "Xcode"))
```
an-tao, line 116: Can we just remove the check of IDE and set all headers as a part of TRANTOR_SOURCES?

Greisby: Well, I just wanted not to touch the project when not generating for non-IDE, since they are not needed then.
But it you want, I'll change it today.

an-tao: @marty1885 What do u think about this?

marty1885: I don't get why the detection is needed. (My guess is they didn't show up in the IDEs' source menu?). 

I agree, let's make the build script simpler.

Greisby:
Strange that you have to re-apply clang-format each time.
Seems that VS2019 does not respect the rules.

marty1885:
The `format.sh` script does more than just clang-foramt. It also converts CRLF (windows line end) into LF (UNIX line end)

trantor/utils/Utilities.h
```diff
@@ -0,0 +1,117 @@
...
+ * @param str String to convert
+ *
+ * @return converted string.
+ */
+TRANTOR_EXPORT std::wstring from_utf8(const std::string &str);
+
+/**
+ * @details Convert a wide string path with arbitrary directory separators
+ * to a UTF-8 portable path for use with trantor.
+ *
```
marty1885, line 46: Although I'm a supporter of snake_case. Please use Camel case since that's the coding style.

Greisby: Oh yes, you're right.
I didn't take care when I initially made them in drogon.
And when I moved them to trantor, I didn't thought about renaming them.
I'll quickly do that.

Greisby:
The lint checking is annoying.
Sometimes it wants 2 spaces after *#else* and sometimes 3 spaces.

marty1885 (approved):
LGTM

PR #159, opened 2021-07-16
Add BUILD_DOC cmake option

Greisby:
Hi,
If you don't mind, I added to CMakeLists.txt the **BUILD_DOC** option and target rules to build and install the Doxygen documentation.
Tested to work on Windows & OS X.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.5)
 project(trantor)
 
-option(BUILD_DOC "Build Doxygen documentation" FALSE)
+option(BUILD_DOC "Build Doxygen documentation" OFF)
 
 list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake_modules/)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -258,3 +260,36 @@ install(EXPORT TrantorTargets
         DESTINATION "${INSTALL_TRANTOR_CMAKE_DIR}"
         NAMESPACE Trantor::
         COMPONENT dev)
+
+# Doxygen documentation
+find_package(Doxygen OPTIONAL_COMPONENTS dot dia)
+if(DOXYGEN_FOUND)
+  set(DOXYGEN_PROJECT_BRIEF "Non-blocking I/O cross-platform TCP network library, using C++14")
+  set(DOXYGEN_OUTPUT_DIRECTORY docs/${PROJECT_NAME})
```
an-tao, line 268: Should we add the `docs` to the .gitignore?

Greisby: No, it is generated in ${CMAKE_CURRENT_BINARY_DIR}

Greisby: Well, if you do not generate in a separate directory, then yes.

PR #164, opened 2021-08-08
Avoid a small memory leak

ljn917:
`SSL_CONF_CTX_free()` needs to be called after `SSL_CONF_CTX_finish()`
OpenSSL docs: https://www.openssl.org/docs/man1.1.1/man3/SSL_CONF_CTX_new.html

I only tested with OpenSSL 1.1.1, but I believe OpenSSL 1.0.x should be the same.


marty1885:
@an-tao I was about the send basically the same patch. And can confirm it works. With this PR drogon's integration test would have no leak.

an-tao:
@ljn917 Thanks so much for your patch

PR #167, opened 2021-08-19
Disable setting SSL Configs when using LibreSSL

marty1885:
LibreSSL (mostly used by OpenBSD) does not support SSL_CONF* which is introduced in #148. This PR disables that support when encountering LibreSSL.

Should this be a warning or a fatal error?

### DIFF
--- a/trantor/net/inner/TcpConnectionImpl.cc
+++ b/trantor/net/inner/TcpConnectionImpl.cc
@@ -204,7 +204,7 @@ class SSLContext
         ctxPtr_ = SSL_CTX_new(TLS_method());
         if (sslConfCmds.size() != 0)
         {
-            LOF_WARN << "LibreSSL does not support SSL confuration commands";
+            LOF_WARN << "LibreSSL does not support SSL configuration commands";
         }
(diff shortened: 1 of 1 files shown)


trantor/net/inner/TcpConnectionImpl.cc
```diff
@@ -204,8 +204,7 @@ class SSLContext
         ctxPtr_ = SSL_CTX_new(TLS_method());
         if (sslConfCmds.size() != 0)
         {
-            LOG_FATAL << "LibreSSL does not support SSL confuration commands";
-            exit(1);
+            LOF_WARN << "LibreSSL does not support SSL confuration commands";
```
an-tao, line 207: LOG_WARN configuration

marty1885: So the conclusion is to use FATAL?

an-tao: I mean there are some typos

marty1885: oops

PR #168, opened 2021-08-23
Enable checking for revoked certifcate

marty1885:
This PR enables trantor's TCP client to check is the server's certificate is revoked by the CA. This _may_ cause incompatibilities 
if the user supplied their own certificate store and the stores doesn't have a CLR (certificate revoke list). Which should be rare and the solution is to add a CLR themselves.


marty1885:
Something is wrong. It's blacklisting way too many sites.

marty1885:
Closing in favor of OCSP. CRLs are deprecated by some CAs.

PR #171, opened 2021-09-01
Improve Error handling for certificates/private keys

neilcook:
Currently the error handling attempts to use errno to print an error, which is mostly useless, because an OpenSSL error parsing the certificate file for example will not set errno. This PR uses the OpenSSL error functions instead to provide a more readable error message, and also includes information about what is actually being processed at the time (i.e. the cert file or the private key etc.)


an-tao:
@neilcook Thank you so much for your patch!

neilcook:
You're welcome, thanks so much for drogon, it's awesome!

an-tao:
> You're welcome, thanks so much for drogon, it's awesome!

Thanks :) and welcome to contribute to Drogon～

PR #176, opened 2021-09-09
Make c-ares support optional when building

neilcook:
Currently c-ares support is always compiled in when it is found,
however it may not be required or needed. This commit makes c-ares support
optional so that it can be disabled at build time, thus making it
a decision for the builder.


an-tao:
@neilcook thanks so much.

PR #193, opened 2021-12-19
Move EventLoop::runAfter to a template

marty1885:
This PR moves `EventLoop::runAfter` into a template. This is a slight optimization that prevents a heap allocation when runAfter is called on the event thread. The old API forces creation of a std::function object no matter what.

### DIFF
--- a/trantor/net/EventLoop.h
+++ b/trantor/net/EventLoop.h
@@ -122,18 +122,6 @@ class TRANTOR_EXPORT EventLoop : NonCopyable
      * f is executed directly before the method exiting.
      */
     template <typename Functor>
-    inline void runInLoop(const Functor &f)
-    {
-        if (isInLoopThread())
-        {
-            f();
-        }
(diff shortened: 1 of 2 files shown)


trantor/net/EventLoop.h
```diff
@@ -121,8 +121,30 @@ class TRANTOR_EXPORT EventLoop : NonCopyable
...
+    {
+        if (isInLoopThread())
+        {
+            f();
+        }
+        else
+        {
+            queueInLoop(f);
+        }
+    }
```
an-tao, line 135: This is not needed because the following universal template can handle this case correctly.

an-tao (changes requested):

trantor/net/EventLoop.h
marty1885: Right. I forgot Functor&& is not a r-valule template reference,

PR #194, opened 2021-12-31
Experimental io_uring support

marty1885:
This PR is still far from complete. But I want to make a PR to keep track of oy progress. And allow someone else to pick up my work in case I don't have time to finish it.

Current progress:

* Adds a IoUringPoller that still works
   * Allow epoll to notify io_uring when events on epoll are ready
   * it does not have any performance benifit **yet**
   * Need to hook up io_uring for IO
* Adds cmake detection for liburing
* 
System requirements:
* Running Linux >= 5.6

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ cmake-build-debug
 .vscode
 .vs
 CMakeSettings.json
+.cache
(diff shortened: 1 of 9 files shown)


uarif1:
Hi @marty1885,

I was interested in the performance affect of io_uring in drogon and I guess this is the change that would be needed to test that.

Just wanted to check if this work was still progressing?

Also was curious why it didnt have any performance benifit?

Thanks!

marty1885:
@uarif1 There's some progress that I haven't pushed. Basically fixing DB connectivity and add proper async IO. However, SSL isn't working and we need to port the current model into using BIO. The channel abstraction that Trantor uses is also broken. That also needs to be fixed to keep the codebase sane.

PR #196, opened 2022-01-13
Add support for sending data streams via callback (TcpConnection::sendStream())

Greisby:
Hi,
This PR adds TcpConnection::sendStream() to send data obtained from a callback.
The goal is to be able to stream data of unknown size with drogon, like transcoded multimedia for example.
sendStream() allows to send chunked data, when drogon sends a response with no Content-Length and the connection is keep-alive.

It is needed for the [PR 1156](https://github.com/drogonframework/drogon/pull/1156) of drogon.

It was tested on Windows 10 (MSVC 2019, x64) and it compiles on OSX (11.5.2).

The callback is a std::function with signature (char*, size_t) -> size_t
At each call, it fills the buffer and returns the size put in the buffer.
The stream ends when 0 is returned.
When the transmission ends, the callback is called once more with a null buffer, so that the callback can cleanup any internals (like allocated memory, file descriptors, transcoder aso.)

### **Note:**
I also did corrections in TcpConnectionImpl.cc on Windows.
`errno` was checked and compared to values of `errno.h`, what's wrong on Windows since Winsocks functions do not set `errno`. The errors have to be retrieved with `WSAGetLastError()`, and the returned values differ from the one of `errno.h`.
What is did is to set `errno` to `WSAGetLastError()`, and redefine the error values (for example `#define EWOULDBLOCK WSAEWOULDBLOCK`).


Greisby:
Note: sendFile() could be rewritten as a wrapper around sendStream(), since it's a particular case of stream, where the size of the stream is known.

an-tao:
@Greisby Thanks so much for your patch, this is a big change :) we need more time to review it.

Greisby:
No pb.
On my side, I’ll work on my project that uses it.
If I find any bugs, I’ll update my PR.

marty1885 (commented):
@Greisby Thanks for the PR! It looks really nice!

May you add a test file? So we can we sure this feature didn't break in future commits. Like those in `trantor/tests`

trantor/net/inner/TcpConnectionImpl.cc
```diff
@@ -602,11 +613,15 @@ void TcpConnectionImpl::readCallback()
...
+            if (errno == EPIPE || errno == ECONNRESET)
             {
-                LOG_DEBUG << "EPIPE or ECONNRESET, errno=" << errno
+#ifdef _WIN32
+                LOG_DEBUG << "WSAENOTCONN or WSAECONNRESET, errno=" << errno
                           << " fd=" << socketPtr_->fd();
+#else
+            LOG_DEBUG << "EPIPE or ECONNRESET, errno=" << errno
+                      << " fd=" << socketPtr_->fd();
+#endif
```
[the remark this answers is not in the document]
an-tao: it seems the clang-format issue. @marty1885

trantor/net/inner/TcpConnectionImpl.cc
```diff
@@ -1461,16 +1497,74 @@ void TcpConnectionImpl::sendFile(FILE *fp, size_t offset, size_t length)
...
 
+void TcpConnectionImpl::sendStream(
+    std::function<std::size_t(char *, std::size_t)> callback,
+    bool chunked)
+{
+    BufferNodePtr node = std::make_shared<BufferNode>();
+    node->offset_ =
+        0;  // node: range offset, length could also be handled for streams
+    node->fileBytesToSend_ = 1;  // force to > 0 until file sent
+    node->streamCallback_ = callback;
```
an-tao, line 1508: Please use `std::move(callback)`

trantor/net/inner/TcpConnectionImpl.cc
```diff
@@ -1503,12 +1597,118 @@ void TcpConnectionImpl::sendFileInLoop(const BufferNodePtr &filePtr)
...
+                if (filePtr->chunked_)  // Put chunk size + chunk data in buffer
+                {
+                    nData = filePtr->streamCallback_(
+                        fileBufferPtr_->data() +
+                            6,  // reserve 6 for header (0-4000 hex size + \r\n)
+                        fileBufferPtr_->size() -
+                            8);  // reserve header + 2 for trailer (\r\n)
+                    fileBufferPtr_->resize(nData + 6);
+                    fileBufferPtr_->push_back('\r');
+                    fileBufferPtr_->push_back('\n');
```
an-tao, line 1625: The underlying functions should not know the details of the upper-layer protocol, I prefer to implement these details in drogon (and the trunked flag is not needed) 
@marty1885 What do you think?

trantor/net/inner/TcpConnectionImpl.cc
```diff
@@ -1461,16 +1497,74 @@ void TcpConnectionImpl::sendFile(FILE *fp, size_t offset, size_t length)
...
     loop_->assertInLoopThread();
-#ifndef _WIN32
-    assert(filePtr->sendFd_ >= 0);
-#else
-    assert(filePtr->sendFp_);
-#endif
+    assert(filePtr->isFile());
 #ifdef __linux__
-    if (!isEncrypted_)
+    if (!isEncrypted_ && !filePtr->chunked_)
```
an-tao, line 1567: I think here the `callback_` shoud be checked instead of the chunked flag. if the sendStream is called by users with a false trucked flag, the sendfile system call will be called with a invalid fd argument.

Greisby:
Hi, sorry, I’m very busy on an urgent project.
I’ll back to you asap.

trantor/net/inner/TcpConnectionImpl.cc
marty1885: Agree. TcpConnection should have no knowledge of HTTP.

Greisby:
I noticed that lint in the ubuntu build **sometimes** (not systematically) wrongly indents code when there are #ifdef / #else / #endif.
The code between the #else and the #endif is less indented than between #if and #else.
Example:
```--- a/trantor/net/inner/TcpConnectionImpl.cc
+++ b/trantor/net/inner/TcpConnectionImpl.cc
@@ -619,8 +619,8 @@ void TcpConnectionImpl::readCallback()
                 LOG_DEBUG << "WSAENOTCONN or WSAECONNRESET, errno=" << errno
                           << " fd=" << socketPtr_->fd();
 #else
-                LOG_DEBUG << "EPIPE or ECONNRESET, errno=" << errno
-                          << " fd=" << socketPtr_->fd();
+            LOG_DEBUG << "EPIPE or ECONNRESET, errno=" << errno
+                      << " fd=" << socketPtr_->fd();
 #endif
```
That's really strange.

Greisby:
Hi, I also tested the trantor/drogon streaming on Mac (OSX 11.6.4 Big Sur).
I had to add a missing library dependency in the CMakeList.txt
It seems to be needed on all versions of OS X, since it is patched in vcpkg for trantor.

rbugajewski:
> Hi, I also tested the trantor/drogon streaming on Mac (OSX 11.6.4 Big Sur). I had to add a missing library dependency in the CMakeList.txt It seems to be needed on all versions of OS X, since it is patched in vcpkg for trantor.

I wonder what happened to the GitHub diff, as it shows like the whole file has changed, but it doesn’t look this is the case?

PR #198, opened 2022-01-17
Wait for loop to exit in EventLoop destructor.

gschmottlach-xse:
This patch forces the EventLoop destructor to actually wait for the eventloop to exit instead of assuming it exits immediately (and if not) asserting. This is discussed in Issue #1154.

### DIFF
--- a/trantor/net/EventLoop.h
+++ b/trantor/net/EventLoop.h
@@ -297,9 +297,9 @@ class TRANTOR_EXPORT EventLoop : NonCopyable
     void abortNotInLoopThread();
     void wakeup();
     void wakeupRead();
-    bool looping_;
+    volatile bool looping_;
     std::thread::id threadId_;
-    bool quit_;
+    volatile bool quit_;
     std::unique_ptr<Poller> poller_;
 
     ChannelList activeChannels_;
--- a/trantor/net/EventLoop.h
+++ b/trantor/net/EventLoop.h
@@ -28,6 +28,7 @@
 #include <functional>
 #include <chrono>
 #include <limits>
+#include <atomic>
 
 namespace trantor
 {
@@ -297,9 +298,9 @@ class TRANTOR_EXPORT EventLoop : NonCopyable
     void abortNotInLoopThread();
     void wakeup();
     void wakeupRead();
-    volatile bool looping_;
+    std::atomic<bool> looping_;
     std::thread::id threadId_;
-    volatile bool quit_;
+    std::atomic<bool> quit_;
     std::unique_ptr<Poller> poller_;
 
     ChannelList activeChannels_;
--- a/trantor/net/EventLoop.h
+++ b/trantor/net/EventLoop.h
@@ -271,7 +271,8 @@ class TRANTOR_EXPORT EventLoop : NonCopyable
      */
     bool isRunning()
     {
-        return looping_ && (!quit_);
+        return looping_.load(std::memory_order_acquire) &&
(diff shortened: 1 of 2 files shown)


trantor/net/EventLoop.h
```diff
@@ -297,9 +297,9 @@ class TRANTOR_EXPORT EventLoop : NonCopyable
     void abortNotInLoopThread();
     void wakeup();
     void wakeupRead();
-    bool looping_;
+    volatile bool looping_;
```
an-tao, line 300: Maybe std::atomic<bool> should be used here.

marty1885: @gschmottlach-xse Please use `std::atomic<bool>`. volatile does not imply atomic. Accessing a volatile variable from multiple threads is still a UB from the compiler's point of view.

timomrs:
Let me mention this here too: this patch, to the best of my understanding, changes the problem I reported in https://github.com/drogonframework/drogon/issues/1157 from assert on exit (reasonably mild) to hang on exit (considerably worse). The correct solution, I believe, is to use a scope guard on `_looping` to make sure it gets set to false when exiting the loop, including when exiting through an exception.

Also,  I think that it would be better to use an `std::condition_variable`(or `atomic::wait`, but that's only since C++20) rather than to semi-busy-wait with a sleep, see for example https://stackoverflow.com/questions/29943775/correct-way-to-wait-on-a-stdatomicint-in-windows/29944112 .

gschmottlach-xse:
You're welcome to address this bug with a PR of your own that addresses this issue in the manner you describe. My solution addressed a problem where the assert was plainly being triggered for a situation that should never occur if the EventLoop was torn down correctly. Basically, you can't just destroy the instance before the event-loop thread has exited. I chose to wait (indefinitely) for the event-loop thread to exit (assuming it would) while you seem to prefer a more abrupt (ab)end. It's a design choice with arguments either way. I don't particularly like waiting (forever) for the event loop thread to exit but the alternative would be to wait some period of time (how long??) before terminating the server. Obviously, the *best* option is to avoid this scenario altogether by making sure the main-loop thread cannot escape looping without informing the rest of the class that it's exited the function. The EventLoop destructor must wait for this to occur. I'd encourage you to submit a PR that addresses both your and my concerns but I'd rather see my PR implemented sooner in Drogon and have your PR introduced when it becomes available. Incremental improvements (from my PR) are arguably better than none at all. It doesn't have to end here if you can contribute additional optimizations.

timomrs:
@gschmottlach-xse sorry, I didn't mean to bash your fix, I certainly see the problem that it addresses. It just happens to make an existing problem (that the loop never exited cleanly if there was an exception) worse (from assert in debug to hang in all settings).

I'd certainly be happy to make a pull request if I had the time, and maybe I eventually will, and I'd certainly base that on top of your PR, which also addresses a real problem. Making the fix itself isn't the issue, that's a couple of lines and maybe 20 minutes of time (to set up the scope guard, the condition variable thing is just a "best practices" suggestion and can be left for yet another PR), but it's setting up the environment to test it.

gschmottlach-xse:
@timomrs - no offense taken. I understand your issue as well but in my situation I was more motivated to have my application exit cleanly so I could detect leaks with Valgrind. I hope you find the time to issue your own PR . . . it would certainly improve things!

rbugajewski:
@timomrs No worries! As @gschmottlach-xse said constant, iterative improvements are better than waiting for the definitive solution. Once there is a working solution for your issue, I'm positive it'll get merged.

PR #201, opened 2022-01-28
Use static members

hwc0919:


hwc0919:
Doesn't fix the problem.

PR #202, opened 2022-01-28
Fix a bug when closing connections on Windows/MacOS

an-tao:


trantor/net/TcpServer.cc
```diff
@@ -162,30 +162,38 @@ void TcpServer::stop()
...
+    if (loop_->isInLoopThread())
+    {
+        handleCloseInLoop(connectionPtr);
+    }
+    else
+    {
+        loop_->queueInLoop(
+            [this, connectionPtr]() { handleCloseInLoop(connectionPtr); });
+    }
+}
```
marty1885, line 196: Maybe just `loop_->runInLoop`?

trantor/net/TcpServer.cc
```diff
@@ -162,30 +162,38 @@ void TcpServer::stop()
...
     else
     {
-        loop_->queueInLoop([this, connectionPtr]() {
-            size_t n = connSet_.erase(connectionPtr);
-            (void)n;
-            assert(n == 1);
+        connLoop->queueInLoop([connectionPtr]() {
             static_cast<TcpConnectionImpl *>(connectionPtr.get())
                 ->connectDestroyed();
         });
```
marty1885, line 181: Same here. This is the same as `connLoop->runInLoop(...)`

PR #208, opened 2022-04-07
Fix tolower with sanitizer cfi

bansan85:
Related to https://github.com/drogonframework/drogon/issues/1209

Using `tolower` in `std::transform` is a violation of cfi sanitizer (Control Flow Integrity). According to https://en.cppreference.com/w/cpp/string/byte/tolower, use `[](unsigned char c){ return std::tolower(c); }`

### DIFF
--- a/trantor/net/TcpClient.cc
+++ b/trantor/net/TcpClient.cc
@@ -238,7 +238,7 @@ void TcpClient::enableSSL(
         std::transform(hostname.begin(),
                        hostname.end(),
                        hostname.begin(),
-                       tolower);
+                       [](unsigned char c) { return tolower(c); });
         SSLHostName_ = std::move(hostname);
     }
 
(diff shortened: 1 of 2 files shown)


marty1885:
Thanks for the PR! 

I don't get what's violating CFI? I checked the diff. We never fed it anything that's not a UTF-8 string, What's the difference between the wrapping a lambda or not?

rbugajewski:
> I don't get what's violating CFI? I checked the diff. We never fed it anything that's not a UTF-8 string, What's the difference between the wrapping a lambda or not?

Yeah, this one looks quite superstitious.

bansan85:
I'm not an expert about cfi violation. I just wanted to run my application built with cfi sanitizer.

Sanitizer complains because I use a `int (int)` function (`tolower`) but std::transform wants a `char(char)`. I just made a try with both code in a simple hello world with godbolt online compiler. None of them generates a crash with cfi sanitizer.

I need to invest more in which condition `tolower` crash and not lambda. Maybe it's related to the version of  stdlibc++ used.

PR #210, opened 2022-04-12
Keep SSL connection half-open when peer EOF

marty1885:
OpenSSL will falsely trigger `SSL_ERROR_SYSCALL` when it tries to read from a socket closed by it's peer ([reference, see the BUGS section](https://www.openssl.org/docs/man1.1.1/man3/SSL_get_error.html)). Trantor doesn't properly handle this edge case and causes the entire socket to be closed upon peer closing the socket.


marty1885:
Close as we switched to a new TLS backend. The approach this PR uses is outdated

PR #215, opened 2022-06-18
Optimization SSL name matching

hope51607:
This PR is aimed at Optimization SSL name matching. 
In the original way it uses Regex to match the SSL name, but it is slow.


marty1885 (approved):
@an-tao I patched a few edge cases and added some testing, Should we good now

PR #216, opened 2022-06-26
Rename BUILD_TRANTOR_SHARED to BUILD_SHARED_LIBS

bansan85:
Use official variable name to build shared library.

You don't need to add static/shared when calling [add_library](https://cmake.org/cmake/help/latest/command/add_library.html) and [CMAKE_POSITION_INDEPENDENT_CODE](https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html#prop_tgt:POSITION_INDEPENDENT_CODE) is automatically set to the good value.


an-tao:
@bansan85 thanks so much for your patch. please update trantor in your PR of drogon.

PR #218, opened 2022-07-15
Add utc methods for trantor::Date

hwc0919:


an-tao:
@hwc0919 Thanks so much. would you please add some unit tests for this patch?

PR #222, opened 2022-09-05
Added SSL Error Trace Log and mTLS Samples

irineu:
Added getOpenSSLErrorStack() method for get detailed SSL error at the log (trace level)
Added ECDH verification (is available)
Added samples with doc to create SSL certs:
- MTLSServer.cc
- MTLSClient.cc


an-tao (commented):
@irineu thanks for your patch. would you please rebase on the master branch?

irineu:
like this?
image

PR #224, opened 2022-09-06
Partial fix of exception safety in the event loop

timomrs:
Previously, if an exception was thrown from the event loop, the whole app would hang in `EventLoop::~EventLoop()`. Fix that by adding a simple scope guard for the loop flag in `EventLoop::loop()`. In addition, catch the exception and allow the `funcsOnQuit`s to run, before rethrowing.

Also superficially fix another similar bug in `doRunInLoopFuncs()`, although as the comments point out, it's exception safety is still under question.

Overall, this exposes another bug whose source is still unclear: if an exception is thrown in a loop function, the program ultimately exits with `FATAL It is forbidden to run loop on threads other than event-loop thread - EventLoop.cc:258. However, this is still better than the previous hang.

Partially addresses [Drogon::1157](https://github.com/drogonframework/drogon/issues/1157)

N.B. We could also run `funcsOnQuit` in the scope guard, though this means that if any function there throws, it'll happen in a destructor while rewinding stack due to an exception, so a double fault and `terminate()` is called immediately. Might nonetheless be the cleaner option?

### DIFF
--- a/trantor/net/EventLoop.cc
+++ b/trantor/net/EventLoop.cc
@@ -182,14 +182,14 @@ namespace
 template <typename F>
 struct ScopeExit
 {
-    ScopeExit(F &&f) : m_f(std::forward<F>(f))
+    ScopeExit(F &&f) : f_(std::forward<F>(f))
     {
     }
     ~ScopeExit()
     {
-        m_f();
+        f_();
     }
-    F m_f;
+    F f_;
 };
 
 template <typename F>
@@ -246,7 +246,7 @@ void EventLoop::loop()
     }
 
     // Run the quit functions even if exceptions were thrown
-    // TOOD: if more exceptions are thrown in the quit functions, some are left
+    // TODO: if more exceptions are thrown in the quit functions, some are left
     // un-run. Can this be made exception safe?
     Func f;
     while (funcsOnQuit_.dequeue(f))
@@ -257,7 +257,7 @@ void EventLoop::loop()
     // Throw the exception from the end
     if (loopException)
     {
-        LOG_WARN << "Rethrowing exception from even loop";
+        LOG_WARN << "Rethrowing exception from event loop";
         std::rethrow_exception(loopException);
     }
 }
--- a/trantor/net/EventLoop.cc
+++ b/trantor/net/EventLoop.cc
@@ -171,29 +171,33 @@ void EventLoop::quit()
     }
 }
 
-
 // The event loop needs a scope exit, so here's the simplest most limited
-// C++14 scope exit available (from https://stackoverflow.com/a/42506763/3173540)
+// C++14 scope exit available (from
+// https://stackoverflow.com/a/42506763/3173540)
 //
-// TODO: If this is needed anywhere else, introduce a proper on_exit from, for example,
-// the GSL library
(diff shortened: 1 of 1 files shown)


timomrs:
it seems the autoformatter fails on the ubuntu build, but actually looking at the output it looks almost as if it's `git diff` that actually errors out (see log). That seems unlikely, so I'm probably just somehow not seeing the actual error? As I don't have experience with clang-format etc, maybe someone can help me out here?

an-tao:
> it seems the autoformatter fails on the ubuntu build, but actually looking at the output it looks almost as if it's `git diff` that actually errors out (see log). That seems unlikely, so I'm probably just somehow not seeing the actual error? As I don't have experience with clang-format etc, maybe someone can help me out here?

let me help you format it with clang-format.

trantor/net/EventLoop.cc
```diff
@@ -170,44 +170,96 @@ void EventLoop::quit()
...
+struct ScopeExit
+{
+    ScopeExit(F &&f) : m_f(std::forward<F>(f))
+    {
+    }
+    ~ScopeExit()
+    {
+        m_f();
+    }
+    F m_f;
```
marty1885, line 192: Please follow the coding style of marking memory variables by append `_` at end of var name. Like `f_` or `functor_`

trantor/net/EventLoop.cc
```diff
@@ -170,44 +170,96 @@ void EventLoop::quit()
...
-    looping_.store(false, std::memory_order_release);
 
+    // Run the quit functions even if exceptions were thrown
+    // TOOD: if more exceptions are thrown in the quit functions, some are left
+    // un-run. Can this be made exception safe?
     Func f;
     while (funcsOnQuit_.dequeue(f))
     {
         f();
     }
```
marty1885, line 255: Minor type. TODO

I think this is fine. Otherwise we run into the problem of needing to handle exceptions recursively. That sounds like a loosing game to me.

marty1885 (commented):
Hi, sorry for the delayed reply. Nice patch! It sovles some of the problem in the library! Really appropriate it. Just some minor style and comments.

trantor/net/EventLoop.cc
timomrs: Fixed typo.

That's a good point about recursive exceptions. C++ itself refuses to do that by forbidding exceptions from destructors when already unwinding an exception.

Which actually makes me think that running these quit functions in the scope exit would be the right alternative: as it is currently stands, if one of the exit functions throws, the original  exception gets lost (except for the log entry that there was an exception). OTOH, if we ran them on the scope exit and an exit function throws, we get terminate, which could be better (though at least in GCC, any info about the previous  exception gets lost anyway). Anyway, I'm leaving this as is for now, and maybe follow  up with another PR later to move them to the scope exit, if you think that would be the right thing?

hwc0919:
Really nice fix!

The hang up problem was introduce in #212, which gave EventLoop static lifetime to protect some late access to it's raw pointer.
It did work, but caused other problems like this one (unhandled exceptions have to wait for destructors of static variables).

Anyway, current design doesn't seem to be enough for safely sharing EventLoop pointer, we have to make some change in the future, maybe in Drogon2.

PR #230, opened 2022-10-08
Fix socket fd leak if Connector destruct before connection callback is made

marty1885:
Currently there is a file descriptor leak in Connector. When the remote host does not create the TCP connection in a timely way. Causing the HTTP client in Drogon to timeout before Connector creates a channel. This can lead to the HTTP client (thus the `TcpClient` and thus the `Connection`, depending on how it's used) to destruct before an IO channel is created. This causes the file descriptor to not be closed. Causing a file descriptor leak

```c++
void Connector::connect()
{
    // this one. Not RAII handled
    int socketfd = Socket::createNonblockingSocketOrDie(serverAddr_.family());
```

This patch fixes this by explicitly checking and closing the created socket if `Connection` destructed before calling any of it's callbacks.

### DIFF
--- a/trantor/net/inner/Connector.cc
+++ b/trantor/net/inner/Connector.cc
@@ -32,9 +32,9 @@ Connector::~Connector()
     if (socketHanded_ == false && fd_ != -1)
     {
 #ifndef _WIN32
-            ::close(fd_);
+        ::close(fd_);
 #else
-            closesocket(fd_);
+        closesocket(fd_);
 #endif
     }
 }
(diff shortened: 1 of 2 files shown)


trantor/net/inner/Connector.cc
```diff
@@ -27,6 +27,18 @@ Connector::Connector(EventLoop *loop, InetAddress &&addr, bool retry)
 {
 }
 
+Connector::~Connector()
+{
+    if (socketHanded_ == false && fd_ != -1)
```
an-tao, line 32: Should this variable be named socketHandled_?

marty1885: I named it `Handed` because the lifetime is handed to IO Channel. We should not attempt to manage it ourselves.

PR #238, opened 2023-02-04
Refactor SSL handling

marty1885:
This PR refactors how Trantor handles SSL. It's much cleaner and supports both Botan and OpenSSL and SSL/TLS provider (in theory other TLS implementations can be used. But maintaining 2 is a big enough task for me).

Before refactor, Trantor uses a bunch of `ifdef` to determine weather SSL can be uses. The refactor uses a layered approach. `BotanTLSProvider` (and `OpenSSLProvider` in the future) is called from `TcpConnectionImpl` (which only deals with raw TCP now). All read and writes to SSL goes through the provider class before being passed to TCP. This makes the code much cleaner and easier to maintain. Also allows alternative SSL implementations like Botan to be used. I imagine I'll be using Botan to add new features faster in the future, then back port them back into OpenSSL

### Status

- Botan TLS backend
  - [x] Client
    - [x] ALPN
    - [x] MTLS
  - [x] Server
    - [x] Only load cert and key once in `enableSSL`
    - [x] ALPN
    - [x] MTLS
- OpenSSL backend
  - [x] Client
    - [x] ALPN
    - [x] MTLS
  - [x] Server
    - [x] Only load cert and key once in `enableSSL`
    - [x] ALPN
    - [x] MTLS
- [x] Expand SSL error handling
  - [x] Add `kSSLProtocolError` to handle misbehaving peer after handshake
- [x] Abstract the certificate class
  - [x] Get SHA-1/SHA-256 fingerprint
  - [x] Get PEM
- [x] sendFile() API support
- [x] sendStream() API support
- [x] Delayed TLS

### Changes/new features

- ALPN support
- API to access basic certificate info from peer
  - SHA1/SHA256 fingerprint
  - PEM encoding of cert
- API to access SNI from server side
- SSL session resume
- Fine grained certificate validation options
  - Date, cert chain, domain name
- Disable weak ciphers on OpenSSL (Botan does not have weak ciphers)
- Make common hashes provided by TLS implementations available via trantor
  - So Drogon doesn't have to know about trantor's TLS backend
  - Added SHA256,SHA3 and BLAKE2b to potentially replace drogon's aging MD5 and SHA1
- Both Botan and OpenSSL supports server/client session caching
- Properly handle close alerts in OpenSSL (trantor does not before)
- Fix memory leak in client destruction
- Good (at least sane) fallback CSPRNG based on Dan Kaminsky's DEFCON talk

### Breaking changes

- `USE_OPENSSL` is no longer a good macro to decide if SSL is supported as Botan is also an option now


trantor/net/inner/BotanTLSConnectionImpl.h
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    std::vector<Botan::X509_Certificate> cert_chain(
+        const std::vector<std::string> &cert_key_types,
+        const std::string &type,
+        const std::string &context) override
+    {
+        // return the certificate chain being sent to the tls client
+        // e.g., the certificate file "botan.randombit.net.crt"
+        return {*cert_};
```
reneme, line 80: You must return your server certificate only if the TLS server implementation specifically asked for a certificate whose subject public key has the type it requested. Otherwise, the certificate gets mis-associated internally and the ciphersuite selection will make mistakes.

Admittedly that's a footgun in the API.

```suggestion
        for(const auto& t : cert_key_types)
        {
            if(t == cert_->subject_public_key_algo().get_oid().to_formatted_string())
            {
                return {*cert_};
            }
        }

        return {};
```

reneme:
I'm happy that you're trying out Botan's TLS implementation for your project. Feel free to drop more feedback on your experiece.

trantor/net/inner/BotanTLSConnectionImpl.h
marty1885: Thank you! This would have took me forever to figure out.
IMO updating the [example code](https://botan.randombit.net/handbook/api_ref/tls.html#code-example) would avoid and steer most people from hitting this landmine.

an-tao (approved):
Excellent works!

CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ project(trantor)
 option(BUILD_DOC "Build Doxygen documentation" OFF)
 option(BUILD_C-ARES "Build C-ARES" ON)
 option(BUILD_SHARED_LIBS "Build trantor as a shared lib" OFF)
+option(TRANTOR_PREFERED_TLS_PROVIDER "Prefered TLS provider for trantor. Valid options are 'openssl', 'botan' or ''" "")
```
an-tao, line 7: TRANTOR_PREFERED_TLS_PROVIDER should be TRANTOR_PREFERRED_TLS_PROVIDER

trantor/net/inner/tlsprovider/BotanTLSProvider.cc
```diff
@@ -0,0 +1,390 @@
...
+        }
+        catch (const Botan::TLS::TLS_Exception &e)
+        {
+            LOG_ERROR << "Unexpected TLS Exception: " << e.what();
+            conn_->shutdown();
+
+            if (tlsConnected_ == false)
+                handleSSLError(SSLError::kSSLHandshakeError);
+            else
+                handleSSLError(SSLError::kSSLProtocolError);
```
reneme, line 166: Maybe inspect the `TLS_Exception` more thoroughly to provide better error categories to your `handleSSLError()`. As described below (in `tls_session_established`):

```suggestion
            if (tlsConnected_ == false) {
                if(e.type() == Botan::Alert::BAD_CERTIFICATE)
                    handleSSLError(SSLError::kSSLInvalidCertificate);
                else
                    handleSSLError(SSLError::kSSLHandshakeError);
            }
            else
                handleSSLError(SSLError::kSSLProtocolError);
```

trantor/net/inner/tlsprovider/BotanTLSProvider.cc
```diff
@@ -0,0 +1,390 @@
...
+        ctx->cert =
+            std::make_unique<Botan::X509_Certificate>(policy.getCertPath());
+    }
+
+    if (!policy.getCaPath().empty())
+    {
+        ctx->certStore = std::make_unique<Botan::Flatfile_Certificate_Store>(
+            policy.getCaPath());
+        if (!server)
+            ctx->requireClientCert = true;
```
reneme, line 379: Just a wild guess, but shouldn't that require client authentication if it is a server? Rather than if it is not?

trantor/net/inner/tlsprovider/BotanTLSProvider.cc
```diff
@@ -0,0 +1,390 @@
...
+
+using namespace trantor;
+using namespace std::placeholders;
+
+static Botan::System_RNG sessionManagerRng;
+// Is this Ok? C++ technically doesn't guarantee static object initialization
+// order.
+static Botan::TLS::Session_Manager_In_Memory sessionManager(sessionManagerRng);
+static thread_local Botan::System_RNG rng;
+static thread_local Botan::System_Certificate_Store certStore;
```
reneme, line 28: I would guess that you don't need to make the `System_Certificate_Store` thread-local. It should be thread-safe on all platforms and by far not contended enough to need a copy per thread.

trantor/net/inner/tlsprovider/BotanTLSProvider.cc
```diff
@@ -0,0 +1,390 @@
...
+        LOG_TRACE << "tls_session_established";
+        bool needCerts =
+            policyPtr_->getValidateDate() || policyPtr_->getValidateDomain();
+        const auto &certs = session.peer_certs();
+        if (needCerts && session.peer_certs().empty())
+        {
+            handleSSLError(SSLError::kSSLInvalidCertificate);
+            throw Botan::TLS::TLS_Exception(Botan::TLS::Alert::NO_CERTIFICATE,
+                                            "No certificate received");
+        }
```
reneme, line 291: I don't think you need to check that here, as the default implementation of `tls_verify_cert_chain` will check that already and throw a `TLS_Exception` if no certificates were sent by the peer. See the other, similar, comments about how to handle this case.

reneme (commented):
I've had a look through your `BotanTLSProvider`. Generally, this looks correct to me. Apart from the way you are trying to handle invalid certificates in `tls_session_established`.

trantor/net/inner/tlsprovider/BotanTLSProvider.cc
```diff
@@ -0,0 +1,390 @@
...
+            channel_ = std::make_unique<Botan::TLS::Client>(
+                *this,
+                sessionManager,
+                *credsPtr_,
+                validationPolicy_,
+                rng,
+                Botan::TLS::Server_Information(policyPtr_->getHostname(),
+                                               conn_->peerAddr().toPort()),
+                policyPtr_->getUseOldTLS()
+                    ? Botan::TLS::Protocol_Version::TLS_V10
```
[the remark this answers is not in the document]
marty1885: I saw this in Botan's issue. I totally agree people should not use TLS 1.0/1.1 and is debating myself if I should enable this. 

Will there be feature test macros to detect support? Or I should just straight up disable this so the API/ABI(?) maintains the same?

marty1885 (commented):
@reneme Thanks! It's very kind of you. 😄

trantor/net/inner/tlsprovider/BotanTLSProvider.cc
```diff
@@ -0,0 +1,390 @@
...
+        {
+            handleSSLError(SSLError::kSSLInvalidCertificate);
+            throw Botan::TLS::TLS_Exception(Botan::TLS::Alert::NO_CERTIFICATE,
+                                            "No certificate received");
+        }
+        if (!certs.empty())
+        {
+            auto &cert = certs[0];
+            if (policyPtr_->getValidateDomain() &&
+                cert.matches_dns_name(policyPtr_->getHostname()) == false)
```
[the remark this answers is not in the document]
marty1885: I've update the handling to use the normal `tls_verify_cert_chain` when possible. However, I can't find a way to allow self signed certificates while still checking the hostname and certificate dates. This is useful for protocols that rely on TOFU.

For example, Gemini uses TOFU but still validates the hostname to detect server misconfiguration and dates to somewhat force renewal.

>  The first time a Gemini client connects to a server, it accepts whatever certificate it is presented. That certificate's fingerprint and expiry date are saved in a persistent database (like the .known_hosts file for SSH), associated with the server's hostname. ....

trantor/net/inner/tlsprovider/BotanTLSProvider.cc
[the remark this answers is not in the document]
reneme: There won't be feature test macros as the TLS 1.0 and 1.1 code was outright removed and won't be supported anymore in the upcoming 3.0 release.
For now, I think it's really up to you whether you want to keep optional support for these protocol revisions in your product.

trantor/net/inner/tlsprovider/BotanTLSProvider.cc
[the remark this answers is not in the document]
reneme: For the self-signed certificate case, you might try using `Botan::PKIX::check_chain()`. That's a low-level API but might serve you well as it has a special case for self-signed end entities. It will check all the things you implemented in `Callbacks::tls_verify_cert_chain()` and more. I personally didn't test it for your use case, though. Please make sure to test it before taking my word for it. ;) Note as well that it won't check any revocation information for the self-signed cert.

Something along those lines should get you started:

```C++
std::shared_ptr<const X509_Certificate> self_signed_ee_cert = /* the certificate */;

std::vector<std::shared_ptr<const X509_Certificate>> self_signed;
self_signed.push_back(self_signed_ee_cert);

Botan::Path_Validation_Restrictions
    restrictions(false /* require revocation */,
                 tls_policy.minimum_signature_strength());

const auto status = 
    Botan::PKIX::check_chain(self_signed, ref_time,
                             hostname, usage,
                             restrictions.minimum_key_strength(),
                             restrictions.trusted_hashes());

const auto result = Botan::PKIX::overall_status(status);

if(result == Botan::Certificate_Status_Code::OK) {
    // success
}
```

Also, I found the policy checks in `Callbacks::tls_verify_cert_chain()` somewhat surprising. Will comment inline shortly.

trantor/net/inner/tlsprovider/BotanTLSProvider.cc
```diff
@@ -0,0 +1,389 @@
...
+    void tls_verify_cert_chain(
+        const std::vector<Botan::X509_Certificate> &certs,
+        const std::vector<std::shared_ptr<const Botan::OCSP::Response>> &ocsp,
+        const std::vector<Botan::Certificate_Store *> &trusted_roots,
+        Botan::Usage_Type usage,
+        const std::string &hostname,
+        const Botan::TLS::Policy &policy)
+    {
+        setSniName(hostname);
+        if (policyPtr_->getValidate() && !policyPtr_->getAllowBrokenChain())
```
reneme, line 308: If `getAllowBrokenChain()` returns `true`, it seems that you will always just do the "reduced" check below ([discussed here](https://github.com/an-tao/trantor/pull/238#discussion_r1128673362) as well). Is that really what you want to do? Personally, I would have expected this policy to enable an additional fallback.

Like so maybe?

```C++
if (policyPtr_->getValidate())
{
   try
   {
      Botan::TLS::Callbacks::tls_verify_cert_chain(...)
   }
   catch (const std::exception&)
   {
      if (!policyPtr_->getAllowBrokenChain())
         throw;

      // perform the reduced check (as sketched in the other comment)
      // and throw if this fails as well
   }
}
```

marty1885: Thanks for pointing our the reduced check. That's a lot of vulnerabilities avoided. I've patched and tested the code to with our unit testes. I'll do more thorough tests in the weekends. 

I don't see how the code in your comment produced different behavior. The try catch logic reduces to the same thing as my if-else. If `getAllowBrokenChain()` is `true`. Then any exception thrown by `Botan::TLS::Callbacks::tls_verify_cert_chain` gets cough and it does the reduced check. And since the reduced check is a subset of the full check. This effectively is the same as just doing the reduced one.

Is there some global state or something I'm missing?

|             | validate | validate + allow broken | no validate |
|-------------|----------|-------------------------|-------------|
| Valid cert  | ok       | ok                      | ok          |
| self signed |          | ok                      | ok          |
| expired     |          |                         | ok          |

reneme: I agree, the final result is the same: Broken chains pass the fallback and therefor the overall check. Fair point. I was confused by the control flow.

As a suggestion: Would it make sense to log a warning if validation is enabled and a broken chain was accepted only due to `getAllowBrokenChain()` returning `true`?

reneme: As a side-note: We're [working on an implementation for RFC 7250](https://github.com/randombit/botan/pull/3355) to (optionally) allow authentication with raw public keys instead of certificate chains. It might take a while until this lands in a public release, but it could be a better fit for TOFU-like use cases than self-signed certs.

trantor/net/inner/tlsprovider/BotanTLSProvider.cc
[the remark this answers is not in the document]
marty1885: I ended up disabling it. No one should use them and I intend on using Botan as the front for new features.

marty1885:
@reneme Thanks! Botan is really a breeze especially after battling with OpenSSL. It's much easier to work with and correct by default. Everything I can ask for from a security  library.

btw, I'm really excited about Botan 3. Looking forward to TLS 1.3 support.

trantor/net/TcpClient.cc
```diff
@@ -66,10 +69,16 @@ TcpClient::TcpClient(EventLoop *loop,
     (void)validateCert_;
     connector_->setNewConnectionCallback(
         std::bind(&TcpClient::newConnection, this, _1));
-    connector_->setErrorCallback([this]() {
-        if (connectionErrorCallback_)
+    // WORKAROUND: somehow we got use-after-free error
+    auto guard = std::shared_ptr<TcpClient>(this, [](TcpClient *) {});
+    auto weakPtr = std::weak_ptr<TcpClient>(shared_from_this());
```
tripleslash, line 74: shared_from_this() can never be called in constructor

trantor/net/TcpClient.cc
```diff
@@ -66,10 +69,16 @@ TcpClient::TcpClient(EventLoop *loop,
     (void)validateCert_;
     connector_->setNewConnectionCallback(
         std::bind(&TcpClient::newConnection, this, _1));
-    connector_->setErrorCallback([this]() {
-        if (connectionErrorCallback_)
+    // WORKAROUND: somehow we got use-after-free error
+    auto guard = std::shared_ptr<TcpClient>(this, [](TcpClient *) {});
```
tripleslash, line 73: this creates a shared_ptr with no relation to the original control block and is illegal

PR #239, opened 2023-02-05
Refactor TcpServer I/O loop logic.

hwc0919:


trantor/net/TcpServer.h
```diff
@@ -89,7 +93,23 @@ class TRANTOR_EXPORT TcpServer : NonCopyable
...
+     * The loops are managed by caller. Caller should ensure that ioLoops
+     * lives longer than TcpServer.
+     *
+     * @param ioLoops
+     */
+    void setIoLoops(const std::vector<trantor::EventLoop *> &ioLoops)
+    {
+        assert(!ioLoops.empty());
+        assert(!started_);
+        ioLoops_ = ioLoops;
```
an-tao, line 112: Here we should reset the loopPoolPtr_ to avoid hold an unused EventThreadPool.

PR #242, opened 2023-03-13
TRNANTOR_LOG_COMPACT - compact logs without source code details(line, file name etc)

ostropik:
- compact logs without source code details(line, file name etc). Preprocessor TRNANTOR_LOG_COMPACT

### DIFF
--- a/trantor/tests/LoggerTest.cc
+++ b/trantor/tests/LoggerTest.cc
@@ -4,6 +4,7 @@
 int main()
 {
     int i;
+    LOG_COMPACT_DEBUG << "Hello, world!";
     LOG_DEBUG << (float)3.14;
     LOG_DEBUG << (const char)'8';
     LOG_DEBUG << &i;
--- a/trantor/utils/Logger.cc
+++ b/trantor/utils/Logger.cc
@@ -205,7 +205,7 @@ Logger::Logger(SourceFile file, int line, bool)
     }
 }
 
-// LOG_COMPACT_TRACE
+// LOG_COMPACT
 Logger::Logger() : level_(kInfo)
 {
     formatTime();
--- a/trantor/utils/Logger.cc
+++ b/trantor/utils/Logger.cc
@@ -245,9 +245,10 @@ RawLogger::~RawLogger()
 }
 Logger::~Logger()
 {
-    if(sourceFile_.data_)
+    if (sourceFile_.data_)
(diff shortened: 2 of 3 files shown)


ostropik:
Can't see gcc errors(
But before I need to understand, does this idea of COMPACT_LOGGER applicable for framework changes, or i should chose another logging framework?

an-tao:
@ostropik Thanks for your patch, IMHO, I prefer to use output Macros such as LOG_COMPACT_TRACE, LOG_COMPACT_DEBUG, etc.
Using the TRANTOR_LOG_COMPACT macro may lead the ABI problem because it appears both in the public header and the implement file.

ostropik:
LOG_COMPACT_TRACE - we can face conflicts with other frameworks which decide also use some LOG_COMPACT_TRACE. So that is why I proposed to use TRNANTOR_LOG_COMPACT. 

For example:
- build trantor lib with this type of logger (for example RELEASE)
- build our specific solution which will use public headers from different libs
- and for use trantor.lib in mode "compact" into huge of preprocessor defines for **different libraries**
- Example of possible collision : some OpenSSL, or MySuperLib decided use LOG_COMPACT_TRACE  but in fact we don't need this

ostropik:
Ok, agree **LOG_COMPACT** . So how can I understand, we can continue with this change request? 

- I yes, plz advise me line/error with GCC. (i have only MVSC)
- If no...so let me know)

an-tao:
> Ok, agree **LOG_COMPACT** . So how can I understand, we can continue with this change request?
> 
> * I yes, plz advise me line/error with GCC. (i have only MVSC)
> * If no...so let me know)

please use clang-format (v11.0.1) to format your code (run the format.sh script). Or I can help you format the code to pass the CI pipeline.

ostropik:
Plz help me with clang-format (v11.0.1). I have tried MSVC plugin, Visual Studio Code, and online formatter.  Can't found mistake exactly in my changes(

an-tao:
You may have misunderstood my meaning. The specific issue I was referring to is that Trantor can be released as a library. If the macro is not defined when compiling Trantor, but is defined when using the Trantor library, it will inevitably lead to compilation errors as the macro affects both the public header files and the source files. We should avoid this design.

ostropik:
Looks like understood you. Please, check the new variant.

ostropik:
Hello, does it need additional action from my side?

PR #253, opened 2023-05-14
Added NetEndian versions of `toIp` and `toIpPort`

Mis1eader-dev:
Adds pure byte versions of toIp and toIpPort, they are compact and use less memory, while also accounting for ipv4 and ipv6.

### DIFF
--- a/trantor/net/InetAddress.cc
+++ b/trantor/net/InetAddress.cc
@@ -9,6 +9,7 @@
 #include <trantor/net/InetAddress.h>
 
 #include <trantor/utils/Logger.h>
+#include <cstring>
 // #include <muduo/net/Endian.h>
 
 #ifdef _WIN32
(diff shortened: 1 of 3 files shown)


an-tao:
@Mis1eader-dev thanks so much for this PR, would you please add some unit tests for this?

Mis1eader-dev:
It's done, can we merge it?

Mis1eader-dev:
There is one potential optimization that could be done for ipv6.
Currently ipv6 addresses are encoded into the string in its full form, meaning "::1" will get encoded as 16 bytes with `toIpNetEndian()`, compared to `toIp()`'s 3 bytes. Meaning the net endian version of `toIp()` is at a disadvantage in encoding ipv6 addresses that have too many zeros.
One way to improve this is to add an additional byte to both ipv4 and ipv6 that indicates the ip version, and for ipv6 to avoid conflict with ipv4 with addresses that have only 4 bytes, although this must be accompanied by a parameter that does this.

Here is why this additional byte is necessary:
The ipv6 address `"::0101:0101"` must be encoded as `std::string({ 1, 1, 1, 1 })`, this would conflict with the ipv4 address `"1.1.1.1"`, that also gets encoded as `std::string({ 1, 1, 1, 1})`.
These conflicts can be avoided by appending one additional byte, `std::string({ 1, 1, 1, 1, 6})` and `std::string({1, 1, 1, 1, 4})`, this still is more compact than `toIp` for a non-printable version of that function, but is not adhering to any standard as far as I am aware.

In my opinion it is worth making this change, this function must only be used in non-printable environments only, such as rate limiting, and user identification.

I will start working on this improvement and add it as a default parameter for both `toIpNetEndian` and `toIpPortNetEndian`.

PR #270, opened 2023-07-06
Fix more build errors on win32/mingw

luk1337:


trantor/utils/crypto/sha256.cc
```diff
@@ -29,7 +29,7 @@
 #define SIG1(x) (ROTRIGHT(x, 17) ^ ROTRIGHT(x, 19) ^ ((x) >> 10))
 
 /**************************** VARIABLES *****************************/
-static const WORD k[64] = {
+static const unsigned int k[64] = {
```
luk1337, line 32: Saw that UINT failed on macos clang, so I changed it to unsigned int.

PR #282, opened 2023-08-14
fix state lifetime

marty1885:


trantor/net/inner/tlsprovider/OpenSSLProvider.cc
```diff
@@ -788,6 +786,8 @@ struct OpenSSLProvider : public TLSProvider, public NonCopyable
     SSL *ssl_;
     BIO *rbio_;
     BIO *wbio_;
+    bool processedHandshakeError_ = false;
```
an-tao, line 789: I prefer to use `{}` in initialized statements.

PR #283, opened 2023-08-16
Fix an error with ssl certificates

an-tao:


tripleslash:
So the server did not send the full certificate chain and instead used to send only the top level certificate? Are you planning a new drogon release with the change?

marty1885:
Can you confirm the `master` branch fixes the issue. We should make a release to address this.

PR #288, opened 2023-08-23
fix OpenSSL read can be incomplete

marty1885:
Fix for https://github.com/drogonframework/drogon/issues/1741

### DIFF
--- a/trantor/net/inner/tlsprovider/OpenSSLProvider.cc
+++ b/trantor/net/inner/tlsprovider/OpenSSLProvider.cc
@@ -720,8 +720,8 @@ struct OpenSSLProvider : public TLSProvider, public NonCopyable
             // horrible syntax, because MSVC
             pending = (std::max)(0, pending);
             recvBuffer_.ensureWritableBytes(
-                (std::min)((std::min)(maxSingleRead, (size_t)pending),
-                size_t{1024}));
+                (std::max)((std::min)(maxSingleRead, (size_t)pending),
+                           size_t{1024}));
             // clamp to int, because that's what SSL_read accepts
             const size_t wrtibleSize =
                 (std::min)(maxWritibleBytes, recvBuffer_.writableBytes());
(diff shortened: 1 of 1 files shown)


trantor/net/inner/tlsprovider/OpenSSLProvider.cc
```diff
@@ -714,15 +714,14 @@ struct OpenSSLProvider : public TLSProvider, public NonCopyable
...
-            auto pending = BIO_pending(inBio);
-            if (pending <= 0)
-                break;
+            auto pending = BIO_pending(rbio_);
             // horrible syntax, because MSVC
+            pending = (std::max)(0, pending);
             recvBuffer_.ensureWritableBytes(
-                (std::min)(maxSingleRead, (size_t)pending));
+                (std::max)((std::min)(maxSingleRead, (size_t)pending),
+                           size_t{1024}));
```
an-tao, line 724: pending = max(1024,pending);
recvBuffer_.ensureWritableBytes((std::min)(maxSingleRead, (size_t)pending));

trantor/net/inner/tlsprovider/OpenSSLProvider.cc
```diff
@@ -714,13 +714,11 @@ struct OpenSSLProvider : public TLSProvider, public NonCopyable
...
         constexpr size_t maxWritibleBytes = (std::numeric_limits<int>::max)();
-        auto inBio = SSL_get_rbio(ssl_);
         while (true)
         {
-            auto pending = BIO_pending(inBio);
-            if (pending <= 0)
-                break;
+            auto pending = BIO_pending(rbio_);
             // horrible syntax, because MSVC
+            pending = (std::min)(1024, pending);
```
an-tao, line 721: Considering the case of pending==0, we need ensureWritableBytes to 1024 bytes, so here should be std::max.

PR #294, opened 2023-09-06
Feature: Integrate spdlog as logging backend for Trantor Logger

Greisby:
Hi,
This PR adds a build option to allow using spdlog as logging backend for trantor::Logger.
I also added a SpdLoggertest for this feature.

The main interest of spdlog integration is that it allows the developers to have trantor & drogon log to whatever they want (file, console, syslog, database aso…), and also unify with the logs of their applications if they use spdlog.

There is a little drawback when enabling spdlog logging and using the trantor::RawLogger:
spdlog logs line by line. Therefore, after every LOG_RAW, in the destructor of RawLogger, a log message will be logged.\
Example:
> LOG_RAW << "This ";
> LOG_RAW << "is a ";
> LOG_RAW << "test\n";

will produce a single line with `This is a text` with spdlog disabled, and 3 lines with spdlog enabled:
> This
> is a
> test

Remark: since I added doxygen documentation, I also fixed misc other doxygen documentation issues.
I hope it's not a problem.


Greisby:
Well, everything seems fine now, but the Windows checks fail building (or running?) the async_file_logger test:
```
Building Custom Rule D:/a/trantor/trantor/trantor/tests/CMakeLists.txt
  AsyncFileLoggerTest.cc
  async_file_logger_test.vcxproj -> D:\a\trantor\trantor\build\trantor\tests\Debug\async_file_logger_test.exe
##[error]Process completed with exit code 1.

```
On my Windows here, it works.
And since cmake doesn't detect spdlog during the configuration on Windows, trantor is anyway built without spdlog support.
So I don't think that it's linked to my PR.

Anyone any idea?

Greisby:
Well, it now builds on Windows also, but spdlog is not detected by the build, although it is installed by conan. I suspect that there is no cmake module file for spdlog in the conan package.

Greisby:
A final remark: I could perhaps remove the `#ifdef TRANTOR_SPDLOG_SUPPORT` in Logger.h, and put it in the functions definitions, that would then be noop functions if not enabled. There would then be no need in drogon to check whether trantor was built with or without spdlog support. Especially when it was not built as a submodule of trantor.

What do you think about it?

an-tao:
> A final remark: I could perhaps remove the `#ifdef TRANTOR_SPDLOG_SUPPORT` around the new function declarations in Logger.h, and put them in their body, that would then be noop when trantor not built with spdlog support. And eventually, I could also add a hasSpdLogSupport() test function. This would avoid potential compilation errors when trantor is not built as a submodule of drogon (like in vcpkg). This way, drogon would only have to check if it finds spdlog at compilation time, and then do a runtime check of trantor capabilities to enable spdlog.
> 
> What do you think about it?

@Greisby I'm sorry for replying so late. I think it's best to remove the TRANTOR_SPDLOG_SUPPORT macro from the public header files.

trantor/utils/Logger.cc
```diff
@@ -243,8 +397,39 @@ RawLogger::~RawLogger()
...
 {
+#ifdef TRANTOR_SPDLOG_SUPPORT
+    static std::map<LogLevel, spdlog::level::level_enum> spdlogLevel{
+        {kTrace, spdlog::level::trace},
+        {kDebug, spdlog::level::debug},
+        {kInfo, spdlog::level::info},
+        {kWarn, spdlog::level::warn},
+        {kError, spdlog::level::err},
+        {kFatal, spdlog::level::critical}};
+    auto spdLogger = getSpdLogger(index_);
```
an-tao, line 411: I think using std::vector or array is more efficient here.

Greisby:
Hello,
Yes, you are totally right. I’ll use a basic array, there is nothing faster :)
Btw, I saw that there is no check on the value of level_ when creating a LogStream object, allowing a potential out-of-bound access to logLevelStr and now also spdLogLevel.
I’ll therefore also add a clamp() when it's initialised.

Greisby:
Stupid me.
Of course there is a better solution than using an array.
A simple cast is enough since they have the same values.

PR #308, opened 2023-12-15
Simplify the code for sending data

an-tao:


tripleslash:
@an-tao If you're doing a rewrite on this, is it possible you design it with support for asynchronous newStreamResponse in mind? Would allow for example server-sent events as mentioned here: https://github.com/drogonframework/drogon/issues/1587

an-tao:
> @an-tao If you're doing a rewrite on this, is it possible you design it with support for asynchronous newStreamResponse in mind? Would allow for example server-sent events as mentioned here: [drogonframework/drogon#1587](https://github.com/drogonframework/drogon/issues/1587)

Yes, I'm considering providing a new API for this. [drogonframework/drogon#1833](https://github.com/drogonframework/drogon/pull/1883), this is a recent solution, it also requires Trantor’s new API.

PR #312, opened 2023-12-17
Remove mutex

an-tao:
1. Simplify the code for sending data；
2. Remove the mutex when sending data;
3. Add an API for sending an async stream;


tripleslash:
The TCP connection timeout needs to be stopped when async stream is open. I made a seperate PR to support this.

PR #321, opened 2024-01-25
Use CryptGenRandom() instead of RtlGenRandom()

luk1337:
32-bit MinGW 8.1.0 doesn't support RtlGenRandom().

### DIFF
--- a/trantor/utils/Utilities.cc
+++ b/trantor/utils/Utilities.cc
@@ -428,7 +428,10 @@ static bool systemRandomBytes(void *ptr, size_t size)
       (__GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 25))))
     return getentropy(ptr, size) != -1;
 #elif defined(_WIN32)  // Windows
-    return RtlGenRandom(ptr, (ULONG)size);
+    static HCRYPTPROV handle = 0;
+    if (!handle && !CryptAcquireContext(&handle, 0, 0, PROV_RSA_FULL, CRYPT_VERIFYCONTEXT | CRYPT_SILENT))
+        return false;
+    return CryptGenRandom(handle, size, (BYTE*)ptr);
 #elif defined(__unix__) || defined(__HAIKU__)
     // fallback to /dev/urandom for other/old UNIX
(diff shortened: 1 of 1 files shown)


trantor/utils/Utilities.cc
```diff
@@ -428,7 +428,10 @@ static bool systemRandomBytes(void *ptr, size_t size)
       (__GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 25))))
     return getentropy(ptr, size) != -1;
 #elif defined(_WIN32)  // Windows
-    return RtlGenRandom(ptr, (ULONG)size);
+    static HCRYPTPROV handle = 0;
+    if (!handle && !CryptAcquireContext(&handle, 0, 0, PROV_RSA_FULL, CRYPT_VERIFYCONTEXT | CRYPT_SILENT))
+        return false;
+    return CryptGenRandom(handle, size, (BYTE*)ptr);
```
an-tao, line 434: Important:  This API is deprecated. New and existing software should start using [Cryptography Next Generation APIs.](https://learn.microsoft.com/en-us/windows/desktop/SecCNG/cng-portal) Microsoft may remove this API in future releases.

The above are the tips about this function on Microsoft's documentation website. Is there a better solution?

luk1337: Actually, maybe I don't need this patch? Somehow, my linking error magically disappeared...

marty1885:
FYI, [according to correspondence with Microsoft](https://bugzilla.mozilla.org/show_bug.cgi?id=504270#c4), `RtlGenRandom` is safe to use. And in fact is preferred over `CryptGetRandom` as it doesn't need to import the entire crypto API on Windows.

trantor/utils/Utilities.cc
marty1885: Also this can be raced. `static HCRYPTPROV handle` either has to be `thread_local` or we need a mutex. It is possible for 2 threads to enter this block of code at the exact same time. The C++ standard guarantees static variable initialization is MT-safe. But the following check is not. It is possible for both threads to find `handle == 0` then both calls `CryptAcquireContext`. Then both writes to `handle`.

Issue #306, opened 2023-12-03
HashUnitTest Hash5 Hash256 Faild


hks2002:
See the result in VS 2022
![image](https://github.com/an-tao/trantor/assets/8808496/2f5a9b6d-ec07-494e-be2c-c577115f39d4)
```
D:\DownLoads\trantor-master\trantor\unittests\HashUnittest.cc(12): error: Expected equality of these values:
  toHexString(md5("hello"))
    Which is: "604E0000FE2E00005D600000D5D70000"
  "5D41402ABC4B2A76B9719D911017C592"

D:\DownLoads\trantor-master\trantor\unittests\HashUnittest.cc(13): error: Expected equality of these values:
  toHexString(md5("trantor"))
    Which is: "37490000DBA1000034D30000ACCA0000"
  "95FC641C9E629D2854B0B60F5A51E1FD"

D:\DownLoads\trantor-master\trantor\unittests\HashUnittest.cc(28): error: Expected equality of these values:
  toHexString(sha256("hello"))
    Which is: "0000A7190000A86500006FA10000B2010000BA3B0000872D00007A1A000001C7"
  "2CF24DBA5FB0A30E26E83B2AC5B9E29E1B161E5C1FA7425E73043362938B9824"

D:\DownLoads\trantor-master\trantor\unittests\HashUnittest.cc(31): error: Expected equality of these values:
  toHexString(sha256("trantor"))
    Which is: "0000C1150000FF010000468B000022340000D1AD00001049000020A20000461D"
  "C72002E712A3BA6D60125D4B3D0B816758FBDCA98F2A892077BD4182E71CF6F5"
```

hks2002:
It is caused by the data type conflict:
In windows type `WORD` is `unsigned short`, it not `unsigned int`. 
`unsigned int` is we wanted.

hks2002:
Using raw data type or type defined in stdint.h would be better. such as `unsigned int` or `uint32_t`.
@marty1885 You can wait for my pull request.
Because I fixed the unittest error code `0xc0000135` or error `MSB3073` if set `BUILD_TESTING=ON` in MSVC also. 
After this, we can turn on `BUILD_TESTING` option in MSVC again.

sultannesro21:
Please send money for me in my bank account =(  01425486129800  )
bank name = awash bank
Name= sultan nesro mohammed

+251925052357

PR #323, opened 2024-02-26
fix data type conflict (#306)

hks2002:


marty1885:
Thanks for fixing it. I've one suggestion. Instead of using `unsigned char`, etc.. can we use fixed size types like `uint8_t`, `uint32_t`, etc..? In case we compile it on a platform where that has a different meaning.

marty1885 (approved):

PR #328, opened 2024-03-14
New cmakelists

hks2002:
After this change:
1. CMakeLists is easy to read/modify.
2. CMakeLists is formated by cmake-format.
3. Build option's values are more deterministic, if not find the libs, it will build faild.
4. Options TLS_PROVIDER support 'none', 'openssl', 'botan-3', 'auto'.
5. Provied some scripts to install all dependencies quickly for different OS.
6. Windows develope become easy(dependencies provide by: vcpkg/conan/cmake_fetchcontent).
7. Test for windows is working now.


hks2002:
@an-tao please review the code and ignore the source code lint error, just a few space issue.

README.md
```diff
@@ -32,6 +32,27 @@ cmake -B build -H.
...
+Trantor provides some building options, you can enable or disable them by setting the corresponding variables to `ON` or `OFF` in the cmake command line, cmake file etc...
+
+| Option name | Description | Default value |
+| :--- | :--- | :--- |
+| BUILD_SHARED_LIBS | Build Trantor as a shared lib |  OFF | 
+| BUILD_TESTING | Build tests |  OFF | 
+| BUILD_DOC | Build Doxygen documentation |  OFF | 
+| BUILD_DEPENDENCIES | Fetch and build dependencies, this allow you use latest or change dependencies version, if you don't want to use system package manager provided | OFF | 
+| TRANTOR_USE_SPDLOG | Allow using the spdlog logging library | ON | 
+| TRANTOR_USE_C-ARES | Allow using C-ARES |  OFF |
```
marty1885, line 46: In the past c-ares is default on. Why the change?

an-tao:
It LGTM

an-tao (approved):

cmake/BuildDependencies.cmake
```diff
@@ -0,0 +1,171 @@
...
+    set_standard_build_output(gtest_main)
+    set_standard_build_output(gmock)
+    set_standard_build_output(gmock_main)
+  endif()
+
+  # spdlog
+  if(TRANTOR_USE_SPDLOG)
+    FetchContent_Declare(fmt URL "https://github.com/fmtlib/fmt/archive/refs/tags/10.2.1.tar.gz")
+    FetchContent_Declare(spdlog URL "https://github.com/gabime/spdlog/archive/refs/tags/v1.13.0.tar.gz")
+    FetchContent_MakeAvailable(fmt spdlog)
```
marty1885, line 69: I've questions regarding how are we going to maintain these. I have a feeling that without a need, these automatic dependency management will get out of sync with what's available and we won't have the main/brain power to maintain these.

deps.macos-12.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# This script is used to install dependencies on MacOS 12.0 or newer
+# botan v3
+brew install googletest spdlog c-ares openssl@3  botan
```
marty1885, line 5: DITTO (and other dep installing scripts), how are we going to keep these from getting out-of-date?

cmake/BuildDependencies.cmake
hks2002: This is mainly for windows, and who don't want to install conan and vcpkg(it not easy to install it in some areas, such china mainland).

Do you know a way to download the latest version? 
Indeed, fetchcontent support GIT_REPOSITORY also, it could keep to latest.
```
FetchContent_Declare(
  googletest
  GIT_REPOSITORY https://github.com/google/googletest.git
  GIT_TAG        703bd9caab50b139428cea1aaff9974ebee5742e # release-1.10.0
)
```
but using GIT might cause git clone interupt issue due to the network problem. especially the big repository, such as openssl.

and using conan needs maintain the libs version also.

an-tao: Trantor does not depend on the fmt library. 
and if we fetch the source code of the famous libraries and build them into trantor, this can lead some version conflicts with the same libraries in OS used by users.

hks2002: spdlog needs fmt libray. if using conan or vcpkg, it will download it and build it automaticly.

hks2002: > Trantor does not depend on the fmt library. and if we fetch the source code of the famous libraries and build them into trantor, this can lead some version conflicts with the same libraries in OS used by users.

I think in Linux/Macos, it not nessesary to using BUILD_DEPENDENCIES(FetchContent) at all.  so BUILD_DEPENDENCIES default value is OFF.
But in windows, fmt is missing in most machines. And if you find the verison conflicts, you can modify the download url to keep version same.

hks2002: > Why don't we use the denpendencies mantained by vcpkg or connan.

Succesfully install vcpkg is not so easy due to git clone interupt. 
conan works fine, It just provide a third way.

an-tao: Users can use vcpkg or canon to manager dependencies on windows.
for fmt, you could install it by vcpkg (or conan), for example:
```shell
.\vcpkg.exe install fmt:x64-windows
```

hks2002: > Users can use vcpkg or canon to manager dependencies on windows. for fmt, you could install it by vcpkg (or conan), for example:
> 
> ```shell
> .\vcpkg.exe install fmt:x64-windows
> ```

I know it. if vcpkg installed successfully.

hks2002: If don't want to use CMAKE FectchContent as an option BUILD_DEPENDENCIES, I can delete these code.

marty1885:
@hks2002 I'll look into it. It seems to be caused by missing ranges support in Mac's clang.

I introduced Botan as a backup option in case OpenSSL screwed up again and gave up another Heartbleed. And as a platform to stop me from over-designing the TLS backend towards OpenSSL. We don't need to chase the latest version on Botan.

hks2002:
> @hks2002 I'll look into it. It seems to be caused by missing ranges support in Mac's clang.
> 
> I introduced Botan as a backup option in case OpenSSL screwed up again and gave up another Heartbleed. And as a platform to stop me from over-designing the TLS backend towards OpenSSL. We don't need to chase the latest version on Botan.

It is noteworthy that botan have different libarary name in `cmake target_link_libraries`, conan using 'botan::botan' while vcpkg using 'Botan::Botan'.

hks2002:
ref #352,

Issue #329, opened 2024-03-31
Fails to build with current c-ares


waywardmonkeys:
when trying to update `c-ares` in `vcpkg` from 1.27 to 1.28, building `trantor` failed due to use of deprecated functions from c-ares 1.28.0.

```
...vcpkg/buildtrees/trantor/src/v1.5.17-2fee79a355.clean/trantor/net/inner/AresResolver.cc:127:5: error: 'ares_gethostbyname' is deprecated: Use ares_getaddrinfo instead [-Werror,-Wdeprecated-declarations]
    ares_gethostbyname(ctx_,
    ^
.../vcpkg/installed/arm64-osx/include/ares.h:592:14: note: 'ares_gethostbyname' has been explicitly marked deprecated here
CARES_EXTERN CARES_DEPRECATED_FOR(ares_getaddrinfo) void ares_gethostbyname(
             ^
.../vcpkg/installed/arm64-osx/include/ares.h:145:22: note: expanded from macro 'CARES_DEPRECATED_FOR'
      __attribute__((deprecated("Use " #f " instead")))
                     ^
1 error generated.
```

ErikTheBerik:
Is there any solution to this? I updated c-ares via brew (1.28.1) and I cannot compile drogon due to this trantor error regarding the use of a deprecated function.

marty1885:
@ErikTheBerik Can you try the latest master? I've merged a patch that fixes it on my local machine. I want to close this issue if it gets fixed.

PR #331, opened 2024-04-03
Fix build on latest c-ares

marty1885:
Fix for #329. This PR replaces the use of `ares_gethostbyname` with the new `ares_getaddrinfo`. I found that we used to only accept IPv4 results. Should we enable IPv6?


an-tao:
We can add ipv6 support in a new PR.

PR #335, opened 2024-05-16
Some spell correction

hks2002:


trantor/utils/Date.h
```diff
@@ -212,26 +212,27 @@ class TRANTOR_EXPORT Date
      *  - "2018-01-01 10:10:25:102414" if the @p fmtStr is "%Y-%m-%d %H:%M:%S"
      *    and the @p showMicroseconds is true
      */
-    std::string toCustomedFormattedString(const std::string &fmtStr,
```
an-tao, line 215: We need to keep the old one with adding a [[deprecated("Use xxx instead")]] attribute and remove it in the future.

an-tao (changes requested):

trantor/utils/Date.h
```diff
@@ -212,26 +212,27 @@ class TRANTOR_EXPORT Date
...
-                                          bool showMicroseconds = false) const;
+    std::string toCustomizedFormattedString(
```
an-tao, line 215: toCustomFormattedString() would be better.

an-tao (changes requested):

PR #338, opened 2024-06-15
Add byte order detection for internal SHA1 implementation for MacOS, POWER, RISC-V and s390

marty1885:
This is a fix for https://github.com/drogonframework/drogon/issues/2065. Currently byte order is SHA1 is not defined on MacOS. This is only a problem if no TLS provider is present.


marty1885:
This PR depends on #340 to be merged first

PR #347, opened 2024-08-06
Keep log level consistency.

hwc0919:


hwc0919:
Other places use LOG_DEBUG.
It affects my application level log collection, so I want to make a change.

https://github.com/an-tao/trantor/blob/79e1ffc59aa1b9c8226c2988fe2e1185e97cd795/trantor/net/inner/TcpConnectionImpl.cc#L881-L891

PR #351, opened 2024-08-29
Add setSocketErrorCallback

fantasy-peak:


fantasy-peak:
The client does not want to exit the program when the file descriptor is not sufficient

PR #355, opened 2024-09-09
Feature: TcpServer hot reload SSL file

lizhizhuanshu:
While the TcpServer is running, reload the SSL from the file. New connections will use the new SSL, while existing connections will remain unchanged.

### DIFF
--- a/trantor/net/TcpServer.cc
+++ b/trantor/net/TcpServer.cc
@@ -239,15 +239,15 @@ void TcpServer::reloadSSL()
 {
     if (loop_->isInLoopThread())
     {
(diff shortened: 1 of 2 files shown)


Mis1eader-dev:
Is the `sslContextPtr_` thread-safe?

lizhizhuanshu:
> Is the `sslContextPtr_` thread-safe?

The sslContextPtr_ of TcpServer is only accessed when loop_ receives a new connection request, after which it will pass the shared_ptr to the new TcpConnection. Therefore, modifying your own sslContextPtr_ in the loop thread is not an issue.

PR #358, opened 2024-09-21
Free leaked memory in ares resolver

omarmohamedkh:
Just added `ares_freeaddrinfo` after using the resolve results.


marty1885 (approved):
Thanks for finding this. LGTM.

PR #359, opened 2024-10-11
Optimize memory usage when sending data on Windows

an-tao:


trantor/net/inner/TcpConnectionImpl.cc
```diff
@@ -728,16 +746,42 @@ ssize_t TcpConnectionImpl::sendNodeInLoop(const BufferNodePtr &nodePtr)
 #ifndef _WIN32
 ssize_t TcpConnectionImpl::writeRaw(const void *buffer, size_t length)
 #else
+static int sendDataWin(int fd, const char *buffer, int length, int sendBufSize)
+{
+    int n = 0;
```
marty1885, line 751: This `ifndef` is really confusing for no good reason. The function deceleration `ssize_t TcpConnectionImpl::writeRaw(const void *buffer, size_t length)` is also repeated once at the end of the else clause. We should clean this up.

PR #367, opened 2025-02-20
Replace ipv4 inet_ntop with a handrolled function

kiroma:
`toIp` is a surprisingly hot function when used in Drogon, accounting for a grand 6% of total runtime when I measured it during my stress tests.
I've tried to make the new function as efficient and consistent as possible, avoiding allocations by abusing the small string optimization. The longest IPv4 address has a total of 15 characters which is just enough to fit in a small libstdc++ string.
To abuse this, I preallocated the string and performed the conversion inside of it with a helper function that does work unconditionally, which achieves really good throughput and low run to run variance.
This speeds up IP to string conversion by about 10x
```
-----------------------------------------------------
Benchmark           Time             CPU   Iterations
-----------------------------------------------------
iptos            9.22 ns         9.22 ns     75759258
ntop             93.3 ns         93.3 ns      7407629
```

### DIFF
--- a/trantor/net/InetAddress.cc
+++ b/trantor/net/InetAddress.cc
@@ -196,16 +196,39 @@ bool InetAddress::isLoopbackIp() const
     return false;
 }
 
+static void byteToChars(std::string::iterator& dst, unsigned char byte)
+{
+    *dst = byte / 100 + '0';
+    dst += byte >= 100;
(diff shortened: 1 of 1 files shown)


rbugajewski:
Thank you for your contribution to improving Drogon’s performance! I appreciate your work on this 🎉

PR #381, opened 2025-11-26
Fix range sendFile() for MinGW

luk1337:
Since toNativePath() returns std::string for MinGW win32, we end up in an infinite loop.


luk1337:
rebased just so CI is green, no other changes have been made.

luk1337:
> rebased just so CI is green, no other changes have been made.

tho it seems like github is 503-ing at random when downloading gtest so too bad, no green CI for me.

PR #400, opened 2026-05-25
Add fetch content for c-ares.

fe-dagostino:
Propose to adopt fetch_content and to simply all cmake files.


fe-dagostino:
it seems that I'm fighting all the time with linters!

fe-dagostino:
to be honest my version was more readable :D

an-tao:
@fe-dagostino Thanks for the PR, this is a good improvement overall.

I think we may want to clarify two things in the CMake logic:

1. Prefer `find_package(c-ares)` first, and only fallback to `FetchContent` if it is not found. This would better support offline builds and system/package-manager based workflows.

2. In the system-c-ares case, we still need to keep `find_dependency(c-ares)` in `TrantorConfig.cmake`, because `Trantor::Trantor` will link against `c-ares::c-ares` transitively, and consumers would otherwise fail at configure time.

So essentially:

* system c-ares → must `find_dependency(c-ares)`
* fetch/internal c-ares → no need to propagate dependency

This keeps both modes working correctly.

fe-dagostino:
Hi @an-tao  you are welcome! thanks to you for the library :D 
unfortunately I don't have visibility about all the projects using the trantor library, so I based the PR mostly on the needs I have, don't taking in account the cmake versions. 
coming back to the PR, yes the approach you are proposing is consistent and allow backward compatibility. I personally don't like to prioritize the packages installed in the system but in this case I guess is the best solution.
I believe that current cmake can be cleaned a bit if you decide to move to a newer version.
Please let me know if there is something I can do to help with current PR.

PR #403, opened 2026-06-29
fix: add missing standard library includes in ConcurrentTaskQueue.h

Lectem:
ConcurrentTaskQueue.h uses `std::atomic_bool`, `std::thread`, `std::mutex`, and `std::condition_variable` without including `<atomic>`, `<thread>`, `<mutex>`, or `<condition_variable>`. It compiles on GCC/Clang with glibc because `<functional>` and `<future>` (from TaskQueue.h) transitively pull them in.

Breaks with Zig cc cross-compiling to glibc 2.17 — the transitive includes aren't there.

This just adds the missing includes and alphabetizes the list.


Lectem:
Looks like the previous CI failures were due to windows-latest changes that have since been fixed.

PR #407, opened 2026-07-21
Fix EPOLLRDHUP/POLLRDHUP mismatch on Linux SPARC/SPARC64

passgat:
Fix a build failure reported by Buildroot autobuild for the bootlin-sparc64-glibc configuration:

  EpollPoller.cc:41:26: error: static assertion failed:
  EPOLLRDHUP != POLLRDHUP

On SPARC/SPARC64, EPOLLRDHUP (0x2000) and POLLRDHUP (0x0800) are different bits, breaking the assumption elsewhere in this file that epoll and poll flags are interchangeable.

Introduce a combined RDHUP value for this architecture and translate it before epoll_ctl() and after epoll_wait(), since revents_ is later tested against POLLRDHUP in Channel::handleEvent() — otherwise half-close detection would silently break at runtime.

This patch is inspired by the equivalent fix in lighttpd: https://redmine.lighttpd.net/issues/3251

### DIFF
--- a/trantor/net/inner/poller/EpollPoller.cc
+++ b/trantor/net/inner/poller/EpollPoller.cc
@@ -22,6 +22,9 @@
 #include <assert.h>
 #include <strings.h>
 #include <iostream>
+#if (defined(__sparc__) || defined(__sparc))
+#define TRANTOR_RDHUP 0x2800 /* EPOLLRDHUP | POLLRDHUP */
+#endif
 #elif defined _WIN32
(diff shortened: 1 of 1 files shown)


passgat:
Just a gentle ping on this PR. 
As an update, this has already been adopted downstream in Buildroot for the drogon package integration: https://github.com/buildroot/buildroot/commit/d38b4915e9437c5492836914c0a0fff12b4f1031

PR #413, opened 2026-08-30
Fix Date::timezoneOffset() returning the 1970 offset instead of the current one

### LINKED ISSUE #401: Date::timezoneOffset return 1h with CEST (UTC+2) instead of 2
Hi,

In UTC+2 (CEST), method Date::timezoneOffset return 3600 instead of 7200 secondes.
tested in redhat9.6 with gcc11

All different implementation failed:

EXPECT_EQ(-trantor::Date(1970, 1, 1).secondsSinceEpoch(), 7200); >= 1.5.26 (6ef8f8f6e22c426e2622d0cd2d1a07deb27fd90c)
EXPECT_EQ(-(trantor::Date::fromDbStringLocal("1970-01-03 00:00:00").secondsSinceEpoch() - 2LL * 3600LL * 24LL), 7200); < 1.5.26 

Thanks !

FujiwaraNaoto:
# Fix `Date::timezoneOffset()` returning the 1970 offset instead of the current one

Fixes #401

## Problem

`Date::timezoneOffset()` cached its result in a `static`, computed from a fixed
point in time:

```cpp
int64_t Date::timezoneOffset()
{
    static int64_t offset = -Date(1970, 1, 1).secondsSinceEpoch();
    return offset;
}
```

`Date(1970, 1, 1)` goes through `mktime()` with `tm_isdst = -1`, which asks the
C library *"what offset applied on 1970-01-01 in this zone?"* — not *"what
offset applies now?"*. The library answers correctly; the bug is that the answer
is then reused as a constant for the lifetime of the process.

A UTC offset is a function of the instant, not a property of a zone. It changes
with daylight saving time, and it changes when a zone alters its standard
offset. Sampling it once at the epoch is wrong in both cases.

In `Europe/Paris` the cached value is `+3600` (CET, since France had no DST in
1970 — it was reintroduced in 1976), so every conversion is an hour off for the
whole CEST half of the year. This is exactly what #401 reports.

### Reproducing it without trantor

This program replicates the constructor and the offset expression, then uses
`mktime()`'s write-back to show what offset each date actually gets:

```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <time.h>

/* Does exactly what trantor's Date(y, m, d) constructor does:
   Date.cc:534-555 -- zero the tm with memset, set tm_isdst = -1, call mktime. */
static long epoch_of(int y, int mo, int d)
{
    struct tm tm;
    memset(&tm, 0, sizeof tm);
    tm.tm_isdst = -1;
    tm.tm_year = y - 1900;
    tm.tm_mon = mo - 1;
    tm.tm_mday = d;
    return (long)mktime(&tm);
}

/* With tm_isdst = -1, mktime writes back the offset and the DST flag that
   actually apply on that date. Peek at them. */
static void show(int y, int mo, int d, const char *label)
{
    struct tm tm;
    memset(&tm, 0, sizeof tm);
    tm.tm_isdst = -1;
    tm.tm_year = y - 1900;
    tm.tm_mon = mo - 1;
    tm.tm_mday = d;
    time_t t = mktime(&tm);
    printf("  %-24s gmtoff=%+6ld isdst=%d  zone=%s\n",
           label, tm.tm_gmtoff, tm.tm_isdst, tm.tm_zone);
    (void)t;
}

int main(void)
{
    printf("TZ=%s\n", getenv("TZ"));
    /* This is the body of Date::timezoneOffset() itself. */
    printf("  trantor timezoneOffset() = -epoch(1970-01-01) = %+ld\n",
           -epoch_of(1970, 1, 1));
    show(1970, 1, 1, "1970-01-01 (used)");
    show(2026, 1, 15, "2026-01-15 (winter)");
    show(2026, 7, 1,  "2026-07-01 (summer)");
    return 0;
}
```

```console
$ gcc -Wall -o tzrepro tzrepro.c
$ TZ=Europe/Paris ./tzrepro
TZ=Europe/Paris
  trantor timezoneOffset() = -epoch(1970-01-01) = +3600
  1970-01-01 (used)        gmtoff= +3600 isdst=0  zone=CET
  2026-01-15 (winter)      gmtoff= +3600 isdst=0  zone=CET
  2026-07-01 (summer)      gmtoff= +7200 isdst=1  zone=CEST
```

`mktime()` is not misbehaving: `+3600` really is the offset that applied on
1970-01-01, and the `isdst=0 zone=CET` line confirms it resolved DST correctly
for that date. The defect is that this answer is then cached and reused for
2026-07-01, where the correct answer is `+7200`.

<details>
<summary>Full run across the affected zones</summary>

```console
$ for z in Europe/Paris America/New_York Asia/Tokyo Europe/Lisbon \
           Europe/Kyiv Asia/Almaty Pacific/Apia America/Santiago; do
      TZ=$z ./tzrepro
  done
TZ=Europe/Paris
  trantor timezoneOffset() = -epoch(1970-01-01) = +3600
  1970-01-01 (used)        gmtoff= +3600 isdst=0  zone=CET
  2026-01-15 (winter)      gmtoff= +3600 isdst=0  zone=CET
  2026-07-01 (summer)      gmtoff= +7200 isdst=1  zone=CEST
TZ=America/New_York
  trantor timezoneOffset() = -epoch(1970-01-01) = -18000
  1970-01-01 (used)        gmtoff=-18000 isdst=0  zone=EST
  2026-01-15 (winter)      gmtoff=-18000 isdst=0  zone=EST
  2026-07-01 (summer)      gmtoff=-14400 isdst=1  zone=EDT
TZ=Asia/Tokyo
  trantor timezoneOffset() = -epoch(1970-01-01) = +32400
  1970-01-01 (used)        gmtoff=+32400 isdst=0  zone=JST
  2026-01-15 (winter)      gmtoff=+32400 isdst=0  zone=JST
  2026-07-01 (summer)      gmtoff=+32400 isdst=0  zone=JST
TZ=Europe/Lisbon
  trantor timezoneOffset() = -epoch(1970-01-01) = +3600
  1970-01-01 (used)        gmtoff= +3600 isdst=0  zone=CET
  2026-01-15 (winter)      gmtoff=    +0 isdst=0  zone=WET
  2026-07-01 (summer)      gmtoff= +3600 isdst=1  zone=WEST
TZ=Europe/Kyiv
  trantor timezoneOffset() = -epoch(1970-01-01) = +10800
  1970-01-01 (used)        gmtoff=+10800 isdst=0  zone=MSK
  2026-01-15 (winter)      gmtoff= +7200 isdst=0  zone=EET
  2026-07-01 (summer)      gmtoff=+10800 isdst=1  zone=EEST
TZ=Asia/Almaty
  trantor timezoneOffset() = -epoch(1970-01-01) = +21600
  1970-01-01 (used)        gmtoff=+21600 isdst=0  zone=+06
  2026-01-15 (winter)      gmtoff=+18000 isdst=0  zone=+05
  2026-07-01 (summer)      gmtoff=+18000 isdst=0  zone=+05
TZ=Pacific/Apia
  trantor timezoneOffset() = -epoch(1970-01-01) = -39600
  1970-01-01 (used)        gmtoff=-39600 isdst=0  zone=-11
  2026-01-15 (winter)      gmtoff=+46800 isdst=0  zone=+13
  2026-07-01 (summer)      gmtoff=+46800 isdst=0  zone=+13
TZ=America/Santiago
  trantor timezoneOffset() = -epoch(1970-01-01) = -10800
  1970-01-01 (used)        gmtoff=-10800 isdst=1  zone=-03
  2026-01-15 (winter)      gmtoff=-10800 isdst=1  zone=-03
  2026-07-01 (summer)      gmtoff=-14400 isdst=0  zone=-04
```

Note `America/Santiago`: `isdst=1` on the *sampled* date, so the constant that
gets cached there is the southern-hemisphere summer offset.

</details>

### Zones affected

Summarising the run above:

| TZ | cached value | 2026 winter | 2026 summer | how it breaks |
|---|---|---|---|---|
| `Europe/Paris` | +3600 | +3600 | **+7200** | off by 1h all summer (#401) |
| `America/New_York` | −18000 | −18000 | **−14400** | off by 1h all summer |
| `Europe/Lisbon` | +3600 | **+0** | **+3600** | CET in 1970, WET today — wrong year-round |
| `Europe/Kyiv` | +10800 | **+7200** | +10800 | MSK in 1970 — wrong year-round |
| `Asia/Almaty` | +21600 | **+18000** | +18000 | standard offset changed — wrong year-round |
| `Pacific/Apia` | −39600 | **+46800** | +46800 | crossed the date line — off by 24h |
| `America/Santiago` | −10800 | −10800 | **−14400** | 1970-01-01 is *inside* southern DST, so the DST offset became the constant |

Zones with no DST and no offset change since 1970 (`Asia/Tokyo`, `UTC`, …)
happen to be correct, which is why this went unnoticed.

### Blast radius

All three internal callers were affected:

- `toDbString()` — `after(-timezoneOffset()).toDbStringLocal()`. Since
  `toDbStringLocal()` formats via `localtime_r()` (a correct, instant-dependent
  offset), this mixed a constant offset with a per-instant one — an asymmetric
  round trip that drifts by the DST amount.
- `fromDbString()` — parsed as local time, then added the constant.
- `fromISOString()` — for strings that carry an explicit offset, the *local*
  zone leaked into a conversion that should not depend on it at all.

### 1. The offset is looked up at the instant being converted

A new overload takes the instant; the existing no-argument form is kept for
source compatibility and now answers for *now* rather than for 1970:

```cpp
static int64_t timezoneOffset(int64_t secondsSinceEpoch);
static int64_t timezoneOffset();  // == timezoneOffset(now)
```

The `static` cache is gone. Besides being the bug, it made the behaviour
untestable: with the offset frozen at first call, changing `TZ` inside a process
had no effect, so no test could exercise a second zone.

### 2. Per-platform implementation

Only APIs that actually exist on each target are used:

| | Linux / macOS / BSD | Windows |
|---|---|---|
| local broken-down time | `localtime_r` | `localtime_s` |
| UTC broken-down time | `gmtime_r` | `gmtime_s` |
| fields → UTC epoch | `timegm` | `_mkgmtime` |
| UTC offset at an instant | `tm_gmtoff` | `timegm(localtime(t)) - t` |

`tm_gmtoff` is a BSD extension present in glibc, musl, macOS and every BSD; the
MSVC CRT has no such member, hence the second form there. That form involves no
DST guessing — `localtime_s()` has already resolved DST for the actual instant —
and it agrees with `tm_gmtoff` exactly, verified across DST zones, a
45-minute-offset zone (`Asia/Kathmandu`), a southern-hemisphere zone
(`America/Santiago`) and a date-line change (`Pacific/Apia`).

### 3. The UTC paths no longer use an offset at all

- `toDbString()` formats through `gmtime_r`/`gmtime_s` directly.
- `fromDbString()` parses the fields and converts them with `timegm`/`_mkgmtime`.
  Adding an offset after a local parse cannot work in principle: which offset
  applies depends on the instant, and that is only known *after* the parse.
- `fromISOString()` with an explicit offset converts through `timegm`/`_mkgmtime`
  minus that offset, so the local zone plays no part.

This removes the possibility of the bug recurring at these call sites rather
than merely correcting the value they are handed. Formatting shared by
`toDbStringLocal()` and `toDbString()` moved into one helper, and the field
parsing shared by `fromDbStringLocal()` and `fromDbString()` into another.

Behaviour that is deliberately unchanged: `Date(year, month, ...)`,
`fromDbStringLocal()` and `fromISOString()` for strings without a time zone
still interpret their input as local time.

## Tests

Six regression tests in `DateUnittest.cc`, built on a `TzGuard` RAII helper that
sets `TZ` and calls `tzset()` (required: `localtime_r()` is not obliged to call
it implicitly) and restores the previous value.

- `timezoneOffsetFollowsDst` — `Europe/Paris`, winter 3600 / summer 7200. This
  is #401.
- `timezoneOffsetFollowsHistoricalChanges` — `Europe/Lisbon`, CET in 1970 vs
  WET/WEST today.
- `timezoneOffsetSouthernHemisphere` — `America/Santiago`.
- `toDbStringIsUtcAcrossDst`
- `dbStringRoundTripAcrossTimezones` — 7 zones × 5 instants.
- `fromISOStringIgnoresLocalTimezone`

Five of the six fail against the current `master` implementation and pass with
this change. The sixth (`dbStringRoundTripAcrossTimezones`) passes either way —
the old code was self-consistently wrong, so round trips held — but it is worth
keeping as an invariant.

Each test first checks, without going through trantor, that the requested zone
actually loaded; if the zoneinfo database is absent the C library silently falls
back to UTC, so the test `GTEST_SKIP`s rather than reporting a false failure.

The tests are `#ifndef _WIN32`. The MSVC CRT only understands the legacy
`tzn[+|-]hh[:mm[:ss]][dzn]` form of `TZ` and always applies the US daylight
saving rules, so IANA zone names — and therefore these expectations — cannot be
expressed there.

### Verification performed

- Full `ctest`: 59/59 passing, under `Asia/Tokyo`, `Europe/Paris`,
  `America/New_York`, `Australia/Sydney` and with `TZ` unset.
- `g++ -std=c++14 -Wall -Wextra`: no warnings.
- `clang-format --dry-run -Werror`: clean.
- Both commits verified independently.

Not verified: the Windows branches (`_mkgmtime`, `gmtime_s`, `localtime_s`)
cannot be compiled in my environment. See the limitations below.

## Limitations

- **Windows CI does not run the unit tests.** `windows-msvc.yml` drops
  `-DBUILD_TESTING=ON` with the comment *"Due to unittest by GTest in windows
  runner will comes out 'error MSB3073'"*, so nothing here is exercised on
  Windows beyond compilation. Re-enabling it is out of scope for this PR but
  would be worth doing.
- **Windows has no historical time zone rules.** Its registry holds only the
  current DST rule (plus a few years of Dynamic DST for some zones), so
  `localtime_s()` applies today's rule to past years. The 1970s `Europe/Paris`
  case cannot be reproduced exactly there without an embedded tzdb. The bug this
  PR fixes — ignoring DST for the *current* date — is fixed on Windows too.
- **MSVC rejects instants outside [1970-01-01, 3000-12-31].** `utcOffsetAt()`
  falls back to the epoch in that case rather than failing. Pre-1970 dates were
  already broken on Windows before this change (`mktime()` returns −1 there), so
  this is not a regression.
- No FreeBSD CI exists, but the BSDs share the `tm_gmtoff` path with Linux and
  macOS, both of which are covered.

## A note on the diff

Part of the diff in `Date.cc` is mechanical and carries no behaviour change: the
two-way platform conditionals now put the Windows branch first (`#ifdef _WIN32`
rather than `#ifndef`), the two split include blocks are merged into one, and
the compound `#if !defined(_WIN32) && (...)` guard becomes `#ifdef`/`#elif`.
That accounts for eight otherwise untouched functions appearing in the diff.


FujiwaraNaoto:
Could you please review it?

